Issue #461, opened 2025-11-13
refactor: Address 754 ESLint warnings from newly enabled rules (Phase 2 of Issue #460)


krishagel:
## Summary
Following the successful resolution of critical ESLint errors in Issue #460, this issue tracks the systematic resolution of the remaining 754 ESLint warnings introduced by the new linting rules (react-perf, unicorn, security plugins). These warnings are not blocking but represent technical debt that should be addressed to improve code quality, performance, and maintainability.

## Context
After completing Phase 1 of Issue #460:
- ✅ All 0 errors resolved
- ✅ All security warnings fixed (12 → 0)  
- ✅ All accessibility warnings fixed (35 → 0)
- ⏳ 754 warnings remain (non-critical)

### 1. React Performance Warnings (494 total) - 65%
**Primary culprits:**
- `react-perf/jsx-no-new-function-as-prop`: 404 warnings
  - Inline arrow functions in JSX props causing unnecessary re-renders
- `react-perf/jsx-no-new-object-as-prop`: 60 warnings
  - Inline object literals in JSX props
- `react-perf/jsx-no-new-array-as-prop`: 21 warnings
  - Inline array literals in JSX props
- `react-perf/jsx-no-jsx-as-prop`: 9 warnings
  - JSX elements passed directly as props

**Impact**: These patterns cause React components to re-render unnecessarily as new references are created on each render.

### 2. Code Complexity Warnings (180 total) - 24%
**Function complexity issues:**
- 89 functions exceed complexity limit of 15 (highest: 75)
- 51 functions exceed line limit of 150 (highest: 500 lines)
- 2 functions exceed parameter limit of 5

**Nesting depth issues:**
- 65 blocks nested too deeply (max depth 4 exceeded)
- 7 nested callbacks exceed limit of 3

**Top offenders by complexity:**
- `POST` handlers: complexity 75, 35, 32
- `updatePromptAction`: complexity 54
- `enhanceCapabilities`: complexity 44
- `createFreshserviceTicketAction`: complexity 41

### 3. Other Warnings (80 total) - 11%
- 2 regex optimization suggestions (unicorn)
- Various minor issues

### Phase 1: React Performance (Priority: High)
- [ ] Extract inline functions to useCallback hooks
- [ ] Extract inline objects/arrays to useMemo hooks or constants
- [ ] Move JSX elements out of props
- [ ] Verify no performance regressions with React DevTools Profiler
- [ ] Document any intentional exceptions

[...]

PR #733, opened 2026-02-04
fix: reduce ESLint warnings 883→284 via React Compiler + Tier 1/2 fixes (#461)

krishagel:
## Summary
- Reduces ESLint warnings from **883 to 284** (67.8% reduction, 0 errors)
- Enables **React Compiler v1.0** for automatic memoization, eliminating all 535 react-perf warnings
- Fixes 45 unicorn auto-fixable warnings, adds config overrides for scripts/tests, and fixes 3 one-off warnings

### Tier 1: Auto-fixable unicorn rules (45 warnings)
Mechanical fixes via `eslint --fix`: `Number.parseInt`, `for-of` loops, `.length > 0` checks, unicode escapes, regex character class sorting across ~20 files.

### Tier 2: Config overrides (31 warnings)
- `scripts/` override: disables `security/detect-non-literal-fs-filename`, complexity rules
- `tests/` override: disables complexity rules

### Tier 3: One-off fixes (3 warnings)
- `jsx-a11y/label-has-associated-control` in api-key-create-dialog.tsx
- `logging/use-error-factories` in nexus conversations route.ts
- `react-hooks/exhaustive-deps` in tools-popover.tsx

### React Compiler enablement (599 warnings → 0)
- Install `babel-plugin-react-compiler@1.0.0`
- Enable `reactCompiler: true` in next.config.mjs
- Re-enable react-compiler eslint rules (previously disabled)
- Disable react-perf rules (now redundant — compiler auto-memoizes)
- Fix 38 compiler violations across 21 files:
  - **setState-in-effect (18):** Wrap setState in `startTransition()`
  - **JSX-in-try/catch (1):** Restructure to return JSX outside try block
  - **Refs-during-render (1):** useState+useEffect pattern for adapter creation
  - **Components-during-render (1):** Move memo components to module level
  - **Variable-before-declaration (1):** Hoist variable extraction
  - **Async-in-useEffect (4):** Inline with cancellation flags

## Test plan
- [x] `npm run lint` — 0 errors, 284 warnings
- [x] `npm run typecheck` — clean
- [x] `npm run build` — succeeds
- [ ] Smoke test Nexus chat, Assistant Architect, Admin Activity pages
- [ ] Verify no visual regressions in detail sheets and tool selectors

### DIFF
--- a/app/(protected)/admin/activity/_components/comparison-detail-sheet.tsx
+++ b/app/(protected)/admin/activity/_components/comparison-detail-sheet.tsx
@@ -44,7 +44,7 @@ export function ComparisonDetailSheet({
       getComparisonDetail(comparison.id).then(result => {
         if (cancelled) return
         if (result.isSuccess && result.data) {
-          startTransition(() => { setDetail(result.data) })
+          setDetail(result.data)
         } else {
           toast({
             variant: "destructive",
@@ -52,7 +52,7 @@ export function ComparisonDetailSheet({
             description: result.message,
           })
         }
-        startTransition(() => { setLoading(false) })
+        setLoading(false)
       })
 
       return () => { cancelled = true }
--- a/app/(protected)/admin/activity/_components/execution-detail-sheet.tsx
+++ b/app/(protected)/admin/activity/_components/execution-detail-sheet.tsx
@@ -56,7 +56,7 @@ export function ExecutionDetailSheet({
       getExecutionDetail(execution.id).then(result => {
         if (cancelled) return
         if (result.isSuccess && result.data) {
-          startTransition(() => { setDetail(result.data) })
+          setDetail(result.data)
         } else {
           toast({
             variant: "destructive",
@@ -64,7 +64,7 @@ export function ExecutionDetailSheet({
             description: result.message,
           })
         }
-        startTransition(() => { setLoading(false) })
+        setLoading(false)
       })
 
       return () => { cancelled = true }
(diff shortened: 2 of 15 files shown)


krishagel:
## Reviewer Feedback Response

### CRITICAL: `startTransition` in async callbacks (ADDRESSED)

Removed `startTransition` wrapping from all async callback contexts across 15 files:

- **`.then()/.catch()/.finally()`** — setState is already deferred by the async context; `startTransition` adds no benefit here
- **`setTimeout` callbacks** — same reasoning (attachment.tsx)
- **`MutationObserver` callbacks** — same reasoning (mermaid-diagram.tsx)

**Important nuance**: `startTransition` is **kept** for all synchronous `setState` calls within `useEffect` bodies. The React Compiler (`babel-plugin-react-compiler` v1.0) enforces this — removing it from synchronous effect code causes 26 lint errors:

> *"Calling setState synchronously within an effect can trigger cascading renders and is not recommended"*

The boundary is:
| Context | `startTransition` needed? |
|---------|--------------------------|
| Synchronous setState in useEffect body | Yes (compiler enforced) |
| setState in `.then()/.catch()/.finally()` | No (async defers it) |
| setState in `setTimeout`/`MutationObserver` | No (async defers it) |

Commit: b740bf0

### MEDIUM: Adapter lazy initializer pattern (NO CHANGE)

This was already explored during the React Compiler enablement work. The suggested pattern:

```tsx
const [adapter] = useState(() => createAssistantArchitectAdapter({...}))
```

**Does not work** because the lazy initializer runs during render, and `createAssistantArchitectAdapter` reads from refs (`inputsRef`, etc.) which are not available during render. The compiler rejects this with a "ref access during render" violation.

The current `useState(null) + useEffect` pattern is the correct approach for this case — the adapter depends on values that are only available after mount.

### Verification
- 0 lint errors (280 warnings — unchanged)
- Typecheck: clean
- Build: success
- Net: -26 lines removed across 15 files

PR #805, opened 2026-02-21
fix(mcp): resolve Canva OAuth redirect and PSD Data protocol version on AWS dev

krishagel:
## Summary

- **Canva OAuth**: `getIssuerUrl()` now checks `AUTH_URL` (NextAuth v5 standard) in addition to `NEXTAUTH_URL` and `NEXT_PUBLIC_APP_URL`. Fixes redirect going to `localhost:3000` on AWS dev where only `AUTH_URL` is set in the ECS task definition.
- **PSD Data MCP**: Patched `@ai-sdk/mcp@1.0.21` to accept MCP protocol version `2025-11-25` (November 2025 stable spec). The SDK only supports up to `2025-06-18` — no newer release available. Patch is safe: the spec is backward compatible and we only extend an allowlist.
- **Dockerfiles**: All three Dockerfiles (`Dockerfile`, `Dockerfile.dev`, `Dockerfile.graviton`) updated to `COPY patches/` before `bun install` so the patch applies inside containers.

## Patch details

See `patches/README.md` for full rationale, safety analysis, and removal instructions. The patch should be removed when `@ai-sdk/mcp` releases a version that natively supports `2025-11-25`.

## Test plan

- [x] PSD Data connector resolves tools successfully in local Docker dev
- [x] `bun run lint` passes (0 errors)
- [x] `bun run typecheck` passes
- [ ] Canva OAuth popup completes on AWS dev after deploy
- [ ] PSD Data connector tools resolve on AWS dev after deploy

### DIFF
--- a/lib/mcp/custom-tools/canva/index.ts
+++ b/lib/mcp/custom-tools/canva/index.ts
@@ -37,9 +37,9 @@ function buildTools(accessToken: string): ToolSet {
 export const canvaProvider: CustomToolProvider = {
   key: "canva",
   urlPatterns: [
-    // Anchored patterns prevent evil-canva.com or canva.com.evil.com from matching.
-    /^https?:\/\/canva\.com/i,
-    /^https?:\/\/[\w-]+\.canva\.com/i,
+    // Anchored: start-of-string prevents evil-canva.com; end-of-host prevents canva.com.evil.com
+    /^https?:\/\/canva\.com(?:[/:?#]|$)/i,
+    /^https?:\/\/[\w-]+\.canva\.com(?:[/:?#]|$)/i,
   ],
   buildTools,
 }


krishagel:

### Summary
- **Critical Issues:** 2
- **High Priority:** 4
- **Suggestions:** 5
- **Positive Practices:** 7

---

**1. SSRF — `tokenEndpointUrl` / `authorizationEndpointUrl` Not Validated Before Fetch**
- **Files:** `app/api/connectors/mcp-auth/callback/route.ts:336`, `app/api/connectors/mcp-auth/initiate/route.ts:143`
- **Problem:** In the pre-registered OAuth flow, `credentials.tokenEndpointUrl` is fetched directly without calling `rejectUnsafeMcpUrl()`. The callback route validates `server.url` but not the credential endpoint URLs. The initiate route builds a redirect URL from `authorizationEndpointUrl` without SSRF validation. The `exchangeRefreshToken` path in `connector-service.ts:637` *does* call `rejectUnsafeMcpUrl` — this inconsistency makes the callback/initiate the weakest links.
- **Fix:** Add `rejectUnsafeMcpUrl(credentials.tokenEndpointUrl)` before the fetch in callback, and `rejectUnsafeMcpUrl(credentials.authorizationEndpointUrl)` before building the auth URL in initiate.
- **Reference:** OWASP SSRF Prevention Cheat Sheet

**2. Encrypted Credentials Returned to Caller via `.returning()`**
- **Files:** `actions/admin/connector.actions.ts:294, 396`
- **Problem:** `createMcpServer` and `updateMcpServer` return `ActionState<SelectNexusMcpServer>`. The `.returning()` call returns the full DB row including `oauthCredentials` JSONB with `encryptedClientSecret`. While this is ciphertext (not plaintext), returning it serves no purpose. `listMcpServers` correctly excludes this field and returns `hasOAuthCredentials: boolean` instead — create/update should follow the same pattern.
- **Fix:** Change return types from `SelectNexusMcpServer` to `McpServerWithStats` and strip `oauthCredentials`/`mcpOauthRegistration` before returning.

---

**3. No Server-Side Validation of OAuth Credential URL Fields**
- **File:** `actions/admin/connector.actions.ts:269-276`
- **Problem:** `authorizationEndpointUrl` and `tokenEndpointUrl` are persisted without validation. An admin (or compromised session) could store an internal URL as `tokenEndpointUrl`, which would be fetched server-side during token exchange (stored SSRF).
- **Fix:** Add a `validateOAuthCredentials()` function that calls `validateMcpUrl()` on both endpoint URLs, and validates `clientId` length.

**4. Brute-Force Cookie Fallback Lacks State Comparison**
- **File:** `app/api/connectors/mcp-auth/callback/route.ts:106-122`
- **Problem:** `findCookieByBruteForce` scans all `mcp_auth_state_*` cookies and returns the first non-expired one when state-based lookup fails. The `oauthState` field is never checked in this path — state-binding is completely absent. While cookies are AES-256-GCM encrypted and TTL'd, concurrent flows could be misbound.
- **Fix:** Consider failing closed when state param is absent, or at minimum document which providers require this fallback.

**5. Cookie `secure` Flag Uses `NODE_ENV` Instead of `ENVIRONMENT`**
- **Files:** `app/api/connectors/mcp-auth/initiate/route.ts:167, 224`
- **Problem:** Per project docs (`docs/learnings/aws/2026-02-18-ecs-node-env-vs-environment.md`), ECS sets `NODE_ENV=production` for all environments. The rest of the codebase uses `ENVIRONMENT` to discriminate. This is inconsistent — not a live vulnerability but a latent risk.
- **Fix:** Align to `process.env.ENVIRONMENT === "prod" || process.env.ENVIRONMENT === "staging"`.

**6. No E2E or Integration Tests**
- **Problem:** The PR introduces a new credential storage path (encrypt → DB → decrypt → fetch) with no test coverage. The `token-encryption.ts` module has `_resetForTesting` making tests straightforward.
- **Fix:** Add unit tests verifying that `createMcpServer` with `oauthCredentials` stores an encrypted (not plaintext) secret, and that `getOAuthCredentials` round-trips correctly.

---

### Suggestions (Consider for Improvement)

**7. Client-Side Only Validation for "Keep Existing Secret"**
- **File:** `connector-form-sheet.tsx:66`
- The "leave blank to keep existing" semantic for clientSecret on edit is enforced only client-side. Consider server-side validation or a comment documenting the intent.

**8. Secrets Manager Cache Not Invalidated on Update**
- **File:** `connector-service.ts:747-800`
- `loadOAuthCredentials` caches for 5 minutes. After admin updates, old Secrets Manager credentials remain active for the TTL. The new inline DB path is always fresh.

**9. No Partial Index for `oauth_credentials IS NOT NULL`**
- **File:** `061-mcp-oauth-credentials.sql`
- `listMcpServers` uses `CASE WHEN oauth_credentials IS NOT NULL`. Low priority with small table sizes.

**10. GDPR Data Portability Gap**
- No data export endpoint for `nexus_mcp_user_tokens`. Pre-existing gap, not introduced here.

**11. FERPA Audit Log Consideration**
- `logToolCall` stores tool `input`/`output` in `nexus_mcp_audit_logs`. If MCP tools return student data (e.g., a SIS connector), that data lands unencrypted in the audit table. Consider field-level encryption if expanding to student-facing connectors.

---

- `listMcpServers` uses explicit column selection and returns `hasOAuthCredentials: boolean` — encrypted payload never reaches list views
- `getAvailableConnectors` + `toMcpConnector()` strip all credential fields from user-facing data
- AES-256-GCM with HKDF-SHA-256 key derivation, random 96-bit IV per call — cryptographically sound
- PKCE (S256) implemented for both pre-registered and MCP-native OAuth flows
- `timingSafeEqual` correctly used for CSRF state comparison
- Cookies are httpOnly, sameSite=lax, path-scoped, AES-256-GCM encrypted, 5-minute TTL
- `sanitizeForLogging` applied at action entry points; credential fields never logged

---

### Required Actions
1. Fix SSRF validation gaps on `tokenEndpointUrl` and `authorizationEndpointUrl` in callback and initiate routes
2. Strip `oauthCredentials` from create/update server action return values
3. Add server-side URL validation for OAuth credential endpoint URLs
4. Align cookie `secure` flag to use `ENVIRONMENT`
5. Run `bun run lint && bun run typecheck` after fixes
6. Add test coverage for the encrypt → store → decrypt round-trip

---
*Automated security review by security-analyst-specialist and data-integrity-guardian agents*

PR #812, opened 2026-02-24
fix: prevent nexus conversation remount on browser tab switch

### LINKED ISSUE #811: bug: Nexus conversation remounts on browser tab switch, losing draft text input
## Description

When a user is on an existing Nexus conversation and switches to a different browser tab, then returns, the entire conversation component tree remounts. This causes a loading spinner to flash and — critically — any draft text in the composer input is lost.

## Steps to Reproduce

1. Open an existing Nexus conversation
2. Type some text in the composer input (do NOT send)
3. Switch to another browser tab
4. Switch back to the Nexus tab
5. Observe: loading spinner flashes, conversation reloads, draft text is gone

## Expected Behavior

Returning to the tab should show the conversation exactly as left, with draft text preserved in the composer input.

## Actual Behavior

The entire conversation component tree unmounts and remounts, displaying a loading spinner briefly. All draft text in the composer is lost.

## Root Cause Analysis

The bug is a chain reaction triggered by NextAuth's default `refetchOnWindowFocus` behavior:

### Chain of Events

1. **`SessionProvider` has `refetchOnWindowFocus={true}` (default)** — `components/utilities/session-provider.tsx:10` renders `<SessionProvider>` with no props, so it uses the default which refetches `/api/auth/session` every time the window regains focus.

2. **Session refetch creates a new object reference** — Even if the session data is identical, React receives a new JavaScript object, triggering downstream re-renders.

3. **`ConversationInitializer` has `session` in its `useEffect` deps** — `app/(protected)/nexus/_components/conversation-initializer.tsx:239`:
   ```typescript
   }, [conversationId, status, session])  // ← session object reference change triggers this
   ```

4. **The effect sets `loading = true`** — Line 198: `startTransition(() => { setLoading(true) })`. When loading is true, the component renders a spinner div instead of `children(messages)` (lines 241-250).

5. **Full component tree unmounts** — `NexusRuntimeWrapper`, `ConversationRuntimeProvider`, `Thread`, and `ComposerPrimitive.Input` all unmount. The assistant-ui runtime holding the draft text is destroyed.

### Secondary Factor

`NexusPageContent` (page.tsx:332) also calls `useSession`, causing additional re-renders on session object change.

### What Was Ruled Out

- No `visibilitychange` event listeners in the codebase
- No SWR/React Query `revalidateOnFocus` config
- No polling intervals on the chat path
- `stableConversationId` pattern is correctly implemented
- No `key={conversationId}` on runtime wrapper (correctly removed per architecture doc)

## Proposed Fix

Two complementary options (both may be applied):

### Option A: Remove `session` from ConversationInitializer deps (targeted)

The `session` object is only needed to check `status` (a primitive string already in deps). Remove `session` from the dependency array:

```typescript
// conversation-initializer.tsx line 239
// Before:
}, [conversationId, status, session])

// After:
}, [conversationId, status])
```

### Option B: Disable `refetchOnWindowFocus` on SessionProvider (root cause)

```typescript
// components/utilities/session-provider.tsx
<SessionProvider refetchOnWindowFocus={false}>
```

This eliminates the root trigger entirely. The app already handles unauthenticated redirects separately.

### Recommendation

Apply **both** fixes:
- Option A prevents the unnecessary effect re-run regardless of what triggers session changes
- Option B eliminates the unnecessary network request on every tab switch

## Affected Files

| File | Line | Issue |
|------|------|-------|
| `app/(protected)/nexus/_components/conversation-initializer.tsx` | 239 | `session` in useEffect deps |
| `components/utilities/session-provider.tsx` | 10 | Missing `refetchOnWindowFocus={false}` |
| `app/(protected)/nexus/page.tsx` | 332 | Secondary `useSession` re-render |

## Acceptance Criteria

- [ ] Switching browser tabs and returning does NOT cause conversation remount
- [ ] Draft text in composer input is preserved across tab switches
- [ ] Authentication still works correctly (redirect on session expiry)
- [ ] No loading spinner flash on tab return
- [ ] Existing conversation functionality unchanged (streaming, history, etc.)
- [ ] Lint and typecheck pass

## Testing Plan

- **Manual**: Open conversation, type draft text, switch tabs, return — text should persist
- **Manual**: Let session expire while on another tab, return — should redirect to login
- **E2E**: Consider adding tab-switch test to nexus test suite

## References

- Architecture doc: `/docs/features/nexus-conversation-architecture.md`
- NextAuth refetchOnWindowFocus: [next-auth#3432](https://github.com/nextauthjs/next-auth/discussions/3432)

krishagel:
## Summary

Fixes #811 — Nexus conversations remount when switching browser tabs, causing draft text loss in the composer input.

## Root Cause

NextAuth's `SessionProvider` refetches the session on window focus by default (`refetchOnWindowFocus={true}`). The new session object reference triggered `ConversationInitializer`'s `useEffect` (which had `session` in its dependency array), setting `loading=true` and unmounting the entire conversation component tree — destroying any draft text in the composer.

## Changes

- **`components/utilities/session-provider.tsx`**: Add `refetchOnWindowFocus={false}` to `SessionProvider` — eliminates the unnecessary session refetch on every tab switch
- **`app/(protected)/nexus/_components/conversation-initializer.tsx`**: Remove `session` object from `useEffect` dependency array, keeping only `status` (a stable primitive string). Removed the redundant `session?.user` guard since NextAuth guarantees user data exists when `status === 'authenticated'`

## Test Plan

- [ ] Open existing Nexus conversation, type draft text, switch tabs, return — text preserved
- [ ] Session expiry still redirects to login correctly
- [ ] No loading spinner flash on tab return
- [ ] Conversation streaming/history still works
- [ ] Lint and typecheck pass (verified)

Closes #811

### DIFF
--- a/app/(protected)/nexus/_components/conversation-initializer.tsx
+++ b/app/(protected)/nexus/_components/conversation-initializer.tsx
@@ -176,8 +176,8 @@ export function ConversationInitializer({
       return
     }
 
-    // Note: NextAuth guarantees session.user exists when status === 'authenticated'.
-    // No additional guard needed — status check above is sufficient.
+    // In this project, our NextAuth configuration ensures session.user is set when status === 'authenticated'.
+    // Given that setup (see auth.ts and Session type augmentation), no additional session guard is needed here.
 
     if (!conversationId) {
       startTransition(() => {
--- a/app/(protected)/nexus/_components/conversation-initializer.tsx
+++ b/app/(protected)/nexus/_components/conversation-initializer.tsx
@@ -158,7 +158,7 @@ export function ConversationInitializer({
 }) {
   const [messages, setMessages] = useState<UIMessage[]>([])
   const [loading, setLoading] = useState(true)
-  const { status, data: session } = useSession()
+  const { status } = useSession()
 
   useEffect(() => {
     // Verify authentication before making API call
@@ -176,15 +176,8 @@ export function ConversationInitializer({
       return
     }
 
-    // Guard against authenticated status with missing user data
-    if (status === 'authenticated' && !session?.user) {
-      log.warn('Authenticated but no user data, skipping conversation load')
-      startTransition(() => {
-        setMessages([])
-        setLoading(false)
-      })
-      return
-    }
+    // Note: NextAuth guarantees session.user exists when status === 'authenticated'.
+    // No additional guard needed — status check above is sufficient.
 
     if (!conversationId) {
       startTransition(() => {
@@ -236,7 +229,7 @@ export function ConversationInitializer({
     return () => {
       abortController.abort()
     }
-  }, [conversationId, status, session])
+  }, [conversationId, status])
 
   if (loading) {
     return (
--- a/components/utilities/session-provider.tsx
+++ b/components/utilities/session-provider.tsx
@@ -7,9 +7,14 @@ export default function AuthSessionProvider({
 }: {
   children: React.ReactNode;
 }) {
-  // refetchOnWindowFocus disabled to prevent unnecessary session object churn on tab switch
-  // (which would trigger useEffect deps that include `status`). A 5-minute background interval
-  // preserves session expiry detection without requiring window focus events.
+  // NOTE: This affects the entire app (wraps RootLayout). The trade-off is accepted because:
+  // 1. API routes enforce auth independently (JWT validation on each request)
+  // 2. Stale session objects were causing Nexus useEffect dep instability (see #811)
+  // 3. The 5-minute poll provides a reasonable expiry detection fallback
+  //
+  // refetchInterval added to compensate: polls every 5 min to detect session expiry
+  // (previously no background polling existed; refetchOnWindowFocus covered this).
+  // Note: each open tab generates one /api/auth/session request per interval.
   return (
     <SessionProvider refetchOnWindowFocus={false} refetchInterval={5 * 60}>
       {children}
--- a/components/utilities/session-provider.tsx
+++ b/components/utilities/session-provider.tsx
@@ -7,5 +7,5 @@ export default function AuthSessionProvider({
 }: {
   children: React.ReactNode;
 }) {
-  return <SessionProvider>{children}</SessionProvider>;
+  return <SessionProvider refetchOnWindowFocus={false}>{children}</SessionProvider>;
 }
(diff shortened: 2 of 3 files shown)


krishagel:

### Summary
- **Critical Issues:** 0
- **High Priority:** 0
- **Suggestions:** 3
- **Compliance:** PASS (GDPR, OWASP, FERPA N/A)

---

### Critical Issues
None.

### High Priority
None.

---

### Suggestions (Optional Improvements)

**1. `components/utilities/session-provider.tsx:10` — Consider adding `refetchInterval` for long-lived sessions**

With `refetchOnWindowFocus={false}` and no `refetchInterval`, an expired/revoked session won't be detected until the next navigation or server-side check. API routes independently validate via `getServerSession()` so no data is exposed, but the UX could silently show 401s on very long sessions.

```typescript
<SessionProvider refetchOnWindowFocus={false} refetchInterval={300}>
  {children}
</SessionProvider>
```

**2. `conversation-initializer.tsx:170` — `unauthenticated` branch renders children with empty state**

Currently returns children with empty messages when unauthenticated. Returning `null` would be more explicit (three upstream auth layers make this a non-issue in practice, but defense-in-depth).

```typescript
if (status === 'unauthenticated') {
  log.warn('User not authenticated, skipping conversation load')
  return null
}
```

**3. `conversation-initializer.tsx:194` — Explicit `credentials: 'same-origin'` on fetch**

The browser default is `same-origin`, so this works correctly. Adding it explicitly documents intent and guards against future refactoring.

```typescript
fetch(`/api/nexus/conversations/${conversationId}/messages`, {
  signal: abortController.signal,
  credentials: 'same-origin',
})
```

---

- **Defense-in-depth auth**: Three independent layers (middleware, nexus layout `getServerSession`, messages API `getServerSession`) — `refetchOnWindowFocus` change does not weaken any of them
- **NextAuth type contract correctly applied**: `status === 'authenticated'` guarantees a truthy session per NextAuth source; removed guard is genuinely redundant and well-documented in the comment
- **AbortController cleanup**: In-flight fetch correctly aborted on dependency change and unmount — no race conditions
- **`status` as stable primitive**: Using `status` (string) instead of `session` (new object reference per refetch) is the correct React dependency pattern
- **No sensitive data in logs**: `log.error` exposes only `conversationId` and error message string

---

### Compliance Scan (data-integrity-guardian)

| Check | Status |
|-------|--------|
| PII Exposure | PASS |
| GDPR Article 32 (encryption) | PASS |
| GDPR Article 17 (right to erasure) | PASS |
| OWASP #07 (session management) | PASS |
| OWASP #03 (injection) | PASS |
| FERPA | N/A |

**APPROVED** — No blocking issues. Safe to merge as-is.

---

PR #845, opened 2026-03-11
fix: #837 — Stop client-side polling on expired sessions, add backoff

### LINKED ISSUE #837: fix: Client-side polling errors on expired sessions and stale Server Action cache
## Summary

Two related client-side error patterns creating ~400 error-level log entries over 2 weeks. Both stem from the client continuing to make requests after the user's context is no longer valid.

### 1. "Failed to fetch" on polling APIs — 341 occurrences

`useExecutionResults` hook (178) and `NotificationProvider` (163) continuously poll their respective APIs. When a user's session expires or their network drops, these produce error-level logs.

```json
{"level":"error","message":"[Client] Failed to fetch execution results","hook":"useExecutionResults","error":"Failed to fetch"}
{"level":"error","message":"[Client] Failed to fetch notifications","component":"NotificationProvider","error":"Failed to fetch"}
```

Some are 401 specifically (27 occurrences):
```json
{"error":"Failed to fetch execution results: 401"}
```

**Impact:** Noisy error logs from idle/expired tabs. No user impact since the session is already dead.

**Fix:**
- Stop polling when session status is unauthenticated (check `useSession().status` before fetching)
- Downgrade from `error` to `warn` for network/auth failures on polling endpoints
- Add exponential backoff on consecutive failures to reduce log volume

### 2. "Failed to find Server Action 'x'" — 65 occurrences

```
Error: Failed to find Server Action "x". This request might be from an older or newer deployment.
```

Occurs after deployments when clients have cached pages referencing old server action IDs. The action ID being literally `"x"` is suspicious — may also be automated scanners/bots probing endpoints.

**Impact:** Users on stale cached pages get errors until they hard refresh. If bots, no user impact.

**Fix:**
- Investigate if `"x"` is a real action ID or scanner probing
- If real: Add a deployment version check that triggers a page reload when mismatch detected
- If bots: Rate-limit or ignore; consider lowering log level

## Acceptance Criteria
- [ ] Polling hooks stop fetching when session is expired/unauthenticated
- [ ] Consecutive fetch failures use exponential backoff
- [ ] Log level for polling fetch failures downgraded to warn
- [ ] Server Action "x" root cause identified (real users vs bots)
- [ ] If real users: deployment version mismatch triggers client reload

krishagel:
## Summary
Implements #837 — Fixes two client-side error patterns producing ~400 error-level log entries over 2 weeks from idle/expired tabs.

### Polling session guards
- **`useExecutionResults`**: Added `useSession()` guard to skip fetches when unauthenticated, handle 401 silently, downgrade log level from `error` to `warn`
- **`NotificationProvider`**: Added session guard to polling interval and SSE connection setup — neither runs when session is unauthenticated

### Exponential backoff
- Both polling hooks use exponential backoff on consecutive failures (2x, 4x, 8x cap)
- Replaced `setInterval` with `setTimeout` chain to support dynamic intervals
- Used `isLoadingRef` (ref) instead of `isLoading` (state) in effect deps to prevent timer churn

### API route fix
- `GET /api/execution-results/recent` was hardcoding `{ status: 500 }` for all errors — now extracts `statusCode` from TypedError (e.g., `AUTH_NO_SESSION` → 401)
- Added NaN guard on `limit` query parameter

### Server Action stale bundle detection
- `useAction` hook detects "Failed to find Server Action" errors from deployment mismatches
- Shows "New version available" toast and auto-reloads after 1.5s
- Root cause: Action ID `"x"` is bot/scanner traffic (real IDs are content hashes like `c3a738a...`), not real users

## Test Plan
- [ ] Typecheck passes (`bun run typecheck`)
- [ ] Lint passes (0 errors)
- [ ] Polling stops when session expires (verify in devtools Network tab)
- [ ] Backoff increases interval on consecutive failures
- [ ] SSE does not connect when unauthenticated
- [ ] Execution results API returns 401 (not 500) for unauthenticated requests
- [ ] Stale server action error triggers reload toast

Closes #837


krishagel:

### Summary
- Critical Issues: 0
- High Priority: 2
- Low Priority: 2 (suggestions)
- Data Integrity: 0 new issues (3 pre-existing patterns noted)

---

### High Priority (Must Fix Before Merge)

**1. Unsafe statusCode cast without range validation**
`app/api/execution-results/recent/route.ts:67-68`

The `statusCode` from a `TypedError` is cast and passed directly to `NextResponse.json()` without validating it's a valid HTTP code (100–599). An invalid value (`0`, `-1`, `NaN`) passed to Node's `http.ServerResponse` can cause unhandled errors leaking stack context.

```typescript
// Fix: validate range after extraction
const rawCode = error instanceof Error && "statusCode" in error
  ? (error as { statusCode: number }).statusCode
  : 500
const statusCode = Number.isInteger(rawCode) && rawCode >= 100 && rawCode <= 599
  ? rawCode
  : 500
```

---

**2. Fragile string-match for stale server action detection**
`lib/hooks/use-action.ts:75`

`message.includes("Failed to find Server Action")` is a substring match against a Next.js internal error string. Two concerns:
- Next.js can change this string in any minor version — silently breaks detection
- Any `ActionState.message` containing that substring would incorrectly trigger `window.location.reload()`

```typescript
// Fix: use startsWith for more precise match, check e instanceof Error
if (e instanceof Error && e.message.startsWith("Failed to find Server Action")) {
```

---

**3. `isLoadingRef.current` not reset on 401 early-return path**
`hooks/use-execution-results.tsx:54-58` and `contexts/notification-context.tsx:56-59`

`isLoadingRef.current` is set to `true` at the start of fetch, but the 401 early-return path returns before `finally` resets it. This leaves the ref stuck as `true`, causing `scheduleNext()` to spin in its else-branch with no delay.

```typescript
if (response.status === 401) {
  setResults([])
  setIsLoading(false)
  isLoadingRef.current = false  // add this
  return
}
```

---

**4. `setTimeout` for reload not cancellable**
`lib/hooks/use-action.ts:82`

The 1.5s reload timer is fire-and-forget. If the component unmounts in that window, the reload fires regardless. Low risk (narrow window, reload is disruptive by design), but document as intentional.

---

### Data Integrity Scan

No new PII/FERPA/GDPR issues introduced by this PR. Pre-existing notes (out of scope for this PR):
- `resultData` (opaque JSONB) returned verbatim — future audit should confirm no AI output/user content stored there
- `errorMessage` returned verbatim — confirm it's always system-generated, not user-supplied
- `NODE_ENV !== "production"` guard in `handleError` exposes full error objects in dev — ensure staging sets `NODE_ENV=production`

Access control: verified server-side auth (`getServerSession()`) enforced before DB call; `userId` scoping prevents cross-user access.

---

- Auth enforced at two layers: server-side (`getServerSession()`) and client-side (`sessionStatus !== 'authenticated'`) — correct separation
- `limit` param protected by NaN guard + `Math.min(..., 50)` cap
- `status` param validated against allowlist before DB query
- SSE backoff with jitter prevents thundering herd
- `consecutiveFailures` as `useRef` avoids render churn
- Log level downgrade from `error` to `warn` is correct for expected polling failures

---

### Required Actions
1. Fix all 4 findings above
2. Run `bun run typecheck` and `bun run lint` after fixes

---

krishagel:



### Severity Summary
- **P3 (Should Fix):** 2 found → 2 resolved
- **Follow-up issues:** 2 created

### Changes Made

**Stale polling chain cancellation** (Copilot, Codex, TS reviewer):
- Added `cancelled` boolean flag to each polling `useEffect` — checked in `scheduleNext()` and in the `setTimeout` callback body
- `.then(scheduleNext)` → `.finally(() => { if (!cancelled) scheduleNext() })` — polling continues after both success and errors, stops reliably on cleanup
- Prevents in-flight `.then()` continuations from spawning new timers after effect cleanup

**State reset when unauthenticated** (Copilot, TS reviewer):
- Added `useEffect` watching `sessionStatus === 'unauthenticated'` in both hooks
- Resets results/notifications to `[]`, `isLoading` to `false`, `error` to `null`, `consecutiveFailures` to `0`
- Prevents stuck loading spinner for signed-out users

**Backoff reset on re-authentication** (Claude reviewer):
- `consecutiveFailures.current = 0` added to the initial-fetch effect before calling fetch
- Users who re-authenticate after a failure streak now start with the base interval, not a backed-off one

**`onError`/`setError` in stale action branch** (Copilot, TS reviewer):
- Added `setError(staleMessage)` and `onError?.(staleMessage)` calls in the stale server action detection path
- Callers that disable buttons or track error state via `onError` now receive this notification

### Follow-up Issues Created
- #846: E2E test coverage for session-expiry polling guards
- #847: Extract shared polling-with-backoff hook (when a 3rd consumer is added)

### Review Agents Used
- Architecture strategist (SOLID compliance)
- Code simplicity reviewer (YAGNI enforcement)
- TypeScript reviewer (full mode)
- Security analyst specialist
- Data integrity guardian

### Testing
- `bun run lint` — 0 errors

Thank you for the thorough reviews!

krishagel:



**Incremental review** — only processed new feedback since Round 1 (2026-03-11T19:40:01Z)

### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Medium):** 1 found → 1 resolved
- **P3 (Minor):** 1 found → 1 resolved

### Changes Made
- Added code comment to `scheduleNext` in both polling hooks explaining the full-interval reschedule trade-off when a fetch is in-flight — makes the intentional latency skew explicit for future readers
- Added `typeof window !== 'undefined'` guard on `window.location.reload()` in `use-action.ts` — idiomatic Next.js SSR safety pattern

### Review Agents Used
- TypeScript reviewer (incremental)

### Testing

Thank you for the thorough follow-up review!

krishagel:



**Incremental review** — only processed new feedback since Round 2 (2026-03-11T20:04:28Z)

### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Medium):** 1 found → 1 resolved, 1 deferred with justification
- **P3 (Minor):** 4 found → 4 resolved

### Changes Made

**Jitter added to exponential backoff** (both polling hooks):
- Formula: `Math.random() * 0.2 + 0.9` — uniform distribution in [0.9, 1.1)
- Matches pattern from `use-document-upload-polling.ts` (existing codebase precedent)

**TypedError type guard aligned with codebase idiom** (`route.ts`):
- Changed `"statusCode" in error` → `"code" in error` (TypedError identifier)
- Cast to `TypedError` directly with `?? 500` fallback — consistent with `error-utils.ts:414`

**`useMemo` → `useRef`** for notification-context logger:
- `useRef(createLogger(...)).current` — semantically correct for stable, non-derived values

**Removed `onError`** from stale server action branch:
- Page reloads in 1.5s — calling `onError` causes callers to briefly flash error UI before reload
- `setError` retained for internal state management

**Comments added** explaining `isLoading=true` during `sessionStatus === 'loading'`:
- Intentional — waiting for NextAuth to resolve auth before acting

### Deferred with justification

**Code duplication** — `use-document-upload-polling.ts` uses structurally different polling (maxAttempts, AbortController, success/error callbacks) — not the same pattern as our backoff loop. Issue #847 remains the correct tracking point for extraction when a true 3rd consumer appears.

### Testing

Thank you for the detailed review!

krishagel:



**Incremental review** — only processed new feedback since Round 3 (2026-03-11T20:16:53Z)

### Severity Summary
- **P3 (Suggestions):** 2 found → 2 resolved
- **Non-blocking observations:** 3 acknowledged (tracked in #846, #847)

### Changes Made:
- **route.ts**: Added missing trailing newline (EOF fix)
- **use-action.ts**: Added Next.js 16.x version comment on stale-bundle string guard; added `warn` log via `client-logger` so stale-bundle detection is observable in production logs; added module-level logger instance
- **use-execution-results.tsx + notification-context.tsx**: Added belt-and-suspenders comment on redundant `isLoadingRef.current = false` in 401 branch, clarifying that `finally` always runs and the assignments are kept for readability, not necessity

### Review Agents Used:
- Incremental feedback analysis (no structural agents re-run on Round 4)

### Testing:
- `bun run typecheck`: 0 errors
- `bun run lint`: 0 errors (335 pre-existing warnings, unchanged)

Ready for re-review. Thank you for the thorough feedback across all four rounds!

PR #855, opened 2026-03-12
refactor: #847 - Extract shared polling-with-backoff hook

### LINKED ISSUE #847: refactor: Extract shared polling-with-backoff pattern from useExecutionResults and NotificationProvider
## Context

Both `hooks/use-execution-results.tsx` and `contexts/notification-context.tsx` contain near-identical exponential backoff polling implementations (~45 lines each):
- `isLoadingRef`, `consecutiveFailures`
- `getInterval()` with 2x/4x/8x cap
- `scheduleNext()` with `cancelled` flag cleanup
- `.finally(scheduleNext)` chain

## Proposed refactor

Extract to `lib/hooks/use-polling-with-backoff.ts`:

```typescript
export function usePollingWithBackoff(
  fn: () => Promise<unknown>,
  options: { baseInterval: number; maxMultiplier?: number }
): void
```

Both hooks become consumers.

## When to do this

When a third consumer is added (avoid premature abstraction with only 2 callers). Flagged by architecture-strategist and code-simplicity-reviewer in PR #845 review.

## Reference
- `hooks/use-execution-results.tsx:101-155`
- `contexts/notification-context.tsx:199-248`
- PR #845

krishagel:
## Summary
Implements #847 — extracts the near-identical exponential backoff polling pattern from `useExecutionResults` and `NotificationProvider` into a shared `usePollingWithBackoff` hook.

## Changes
- New `lib/hooks/use-polling-with-backoff.ts` — reusable hook managing backoff state, jitter, concurrent-call guard, and failure tracking via promise resolution/rejection
- Refactored `hooks/use-execution-results.tsx` — uses shared hook, single `fetchResults` function (no duplication)
- Refactored `contexts/notification-context.tsx` — uses shared hook, fetch functions re-throw for backoff tracking with `.catch(() => {})` at non-polling call sites

## Design decisions
- Hook detects success/failure from promise resolution/rejection (no manual `resetFailures`/`recordFailure` needed during polling)
- `resetFailures()` exposed for session-reset edge case
- Failures auto-reset when `enabled` transitions to `true` (covers re-authentication)
- Manual refresh calls are not guarded by the hook's `isLoadingRef` — concurrent GET requests are harmless

## Test Plan
- [x] `bun run typecheck` passes (0 errors)
- [x] `bun run lint` passes (0 errors, only pre-existing warnings)
- [ ] Manual verification: notifications poll with backoff on failure
- [ ] Manual verification: execution results poll with backoff on failure

Closes #847


krishagel:

### Summary
- Critical Issues: 0
- High Priority: 1
- Suggestions: 3
- Data/Compliance Findings: 0

---

### High Priority (Fix Before Merge)

**File:** `lib/hooks/use-polling-with-backoff.ts:39-91`
**Issue:** `isLoadingRef` not reset on effect cleanup — stuck guard on fast unmount

When the `useEffect` tears down while a fetch is in-flight, `isLoadingRef.current` is left as `true`. On the next mount/re-enable, the polling hook's guard (`if (!isLoadingRef.current)`) sees `true` and defers to the `else` branch, silently skipping the first interval window.

**Fix:**
```typescript
return () => {
  cancelled = true
  clearTimeout(timeoutId)
  isLoadingRef.current = false  // Reset guard so next effect run can poll immediately
}
```

---

### Suggestions

**1. `lib/hooks/use-polling-with-backoff.ts:44-45`** — Failure reset fires on every dep change, not just `enabled` transition

`consecutiveFailures.current = 0` runs whenever `fn`, `baseInterval`, or `maxMultiplier` change — not just when `enabled` goes `false → true`. A future consumer with a non-memoized function ref would silently defeat backoff. Fix: track the transition with a `wasEnabledRef`, or document in JSDoc that `fn` must be stable (i.e. wrapped in `useCallback`).

**2. `notification-context.tsx:55-58` / `use-execution-results.tsx`** — 401 early-return behavior undocumented

The 401 path returns without throwing, so the polling hook treats it as success (no backoff increment). This is correct intent but not commented. Add inline note: _"401 returns without throwing — polling hook treats this as success. Session expiry should not trigger exponential backoff."_

**3. No unit tests for `usePollingWithBackoff`**

The backoff math, jitter bounds, concurrent-call guard, and cancellation semantics have no isolated unit tests. E2E coverage exists but is skipped in most CI contexts. Recommend adding `lib/hooks/__tests__/use-polling-with-backoff.test.ts`.

---

### Data Integrity & Compliance

No findings. This PR is a pure timing/backoff refactor — data-agnostic by design. Auth gates preserved, no new PII logged, FERPA/GDPR surfaces unchanged.

---

- `cancelled` flag checked in both `setTimeout` callback and `.finally()` — correctly closes the race window from `docs/learnings/react-patterns/2026-03-11-polling-then-survives-cleartimeout.md`
- `.finally()` used for next-poll scheduling, ensuring continuation on both success and error paths
- All non-hook call sites append `.catch(() => {})` with explanatory comments
- `sessionStatus` (primitive) used as effect dep instead of `session` object
- `useRef` used for `isLoadingRef` and `consecutiveFailures` to avoid effect re-run churn

---

krishagel:



### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Must Fix):** 2 found → 2 resolved
- **P3 (Style):** 3 found → 3 resolved

### Changes Made

**P2 — Off-by-1 in consecutiveFailures log** (flagged by Claude, Codex): The hook's `.catch()` increments the counter *after* the fetch function re-throws, so reading `consecutiveFailures.current` inside the catch block produced a stale value. Fixed by logging `getConsecutiveFailures() + 1` to reflect the post-failure count.

**P2 — maxMultiplier not validated** (flagged by Copilot): A value of 0 or negative would produce 0ms polling intervals, spinning the event loop. Added `Math.max(1, options.maxMultiplier ?? 8)` to clamp the value.

**P3 — Mutable ref in public API** (flagged by Claude): Changed `consecutiveFailures` (MutableRefObject) to `getConsecutiveFailures()` getter so callers can read the value without the ability to accidentally write to it and break hook invariants.

**P3 — `refreshResults` defined inline in return** (flagged by Claude): Moved to named `const refreshResults = useCallback(...)` above the return, consistent with the rest of the file.

**P3 — Closure ordering documented**: Added comment explaining why `getConsecutiveFailures` is safe to reference inside callbacks defined before the hook call.

### Not addressed (out of scope)

- `immediate` option (Gemini): Consolidating initial fetch into the hook is Observation 4 from our own review — noted as a future enhancement. Doing it now would expand PR scope beyond the refactor intent.
- Initial fetch failures bypassing backoff (Codex P2): This matches the original behavior (initial fetch had its own effect, independent of the polling backoff counter). Consistent with Observation 4.

### CI
All checks passing: Test/Lint/TypeCheck ✓, CodeQL ✓, CDK Validate ✓

Ready for re-review.

krishagel:



**Incremental review** — only processed new feedback since Round 1 (2026-03-12T13:17:24Z)

### Severity Summary
- **P1 (Blocking):** 1 found → 1 resolved
- **P3 (Style):** 1 found → 1 resolved

### Changes Made

**P1 — Circular dep / eslint-disable-line in production code** (flagged by both Claude reviewers): `getConsecutiveFailures()` was exposed from the hook and referenced inside `fetchResults`/`fetchNotifications`, which are defined before the hook call. This required an `eslint-disable-line` suppression. Replaced with an `onFailure?: (consecutiveFailures: number) => void` callback option. The hook calls it after incrementing — the consumer receives the accurate post-increment count with no `+1` arithmetic and no suppression comment. Removed `getConsecutiveFailures` from the hook's public API entirely.

**P2 — `getConsecutiveFailures() + 1` fragile coupling** (flagged by Claude): Same root cause — now resolved by the `onFailure` callback. Backoff depth is logged separately from the error log (one log has the requestId + error message, one has the backoff depth).

**P3 — Repeated `.catch(() => {})` comments** (flagged by Claude nit): Consolidated 5x repeated inline comments into a single top-level note on `fetchNotifications` documenting the re-throw contract.

### Not addressed

- `immediate` option / unifying initial fetch into hook: Observation 4, consistently called out as future enhancement across all rounds. Out of scope for this refactor.
- Dev-mode warning for non-memoized `fn`: Nice-to-have, JSDoc documents the requirement. Out of scope.

### CI
All checks passing.

Ready for re-review.

krishagel:



**Incremental review** — only processed new feedback since Round 2 (2026-03-12T14:14:59Z)

### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Must Fix):** 2 found → 2 resolved
- **P3 (Style):** 2 found → 2 resolved

### Changes Made

**P2 — `onFailure` dep array footgun** (flagged by Claude bot): `onFailure` was in the `useEffect` dep array, meaning an inline arrow function passed by a caller would restart the polling chain on every render. Fixed by wrapping `onFailure` in an internal `onFailureRef` (updated via a sync effect after each render) and removing it from the dep array. JSDoc updated to note that memoization is no longer required.

**P2 — Missing `onFailure` unit tests** (flagged by Claude bot): Added three tests:
- `onFailure` is called with the post-increment count after each failure
- `onFailure` is NOT called on successful polls
- Changing the `onFailure` reference does not restart the polling chain (explicitly validates the ref-wrapping behavior)

**P3 — Manual refresh comment clarity** (flagged by Claude bot): Updated `refreshResults` and `refreshNotifications` call sites to clarify that `.catch(() => {})` is a deliberate design choice — manual refresh failures are not counted toward polling backoff.

**P3/nit — `Math.pow` → `**` operator**: Modernized to ES2016 exponentiation operator for consistency.

### Not addressed

- `consecutiveFailures` count removed from per-request logs (item #3): Observability tradeoff of the refactor. The `onFailure` callback provides backoff depth visibility. Adding it back to per-request logs would require re-exposing hook internals or duplicating state. Noted as acceptable regression.

### CI
All 14 tests passing. Typecheck clean.

Ready for re-review.

krishagel:



**Incremental review** — only processed new feedback since Round 3 (2026-03-12T14:49:45Z)

### Severity Summary
- **P1 (Blocking):** 1 found → 1 resolved
- **P3 (Nits):** 2 found → 2 resolved

### Changes Made

**P1 — onFailureRef sync timing gap** (flagged by Claude bot): `useEffect(() => { onFailureRef.current = onFailure })` runs after the browser has painted, leaving a window where a timer callback could call the stale ref. Fixed with a synchronous render-body assignment `onFailureRef.current = onFailure` — the ref is always current by the time any effect or timer callback fires. This is the React-docs-recommended pattern for event handler refs.

**P3 — Unnecessary `useCallback` on `onFailure` in both consumers** (flagged by Claude bot): Dropped the `useCallback` wrappers in `notification-context.tsx` and `use-execution-results.tsx`. The hook wraps `onFailure` in a ref internally so memoization is not required. Keeping `useCallback` contradicted the JSDoc and could mislead future readers. Added inline comment at each call site explaining why it's omitted.

**P3 — `mockReturnValue(0.5)` in `beforeEach` appears redundant** (flagged by Claude bot): Added comment clarifying this is kept as defensive reset in case a future test temporarily overrides the mock mid-suite.

### CI
All 14 tests passing. Typecheck clean.

Ready for re-review.

PR #841, opened 2026-03-11
fix: #835 — SSE streaming validation failures, credential errors, and log noise

krishagel:
## Summary
Fixes three related issues in the Assistant Architect streaming pipeline causing 1,800+ log entries over 2 weeks.

### Changes

**1. Add missing SSE event types** (tool-input-delta, tool-input-available)
- Added TypeScript interfaces, Zod validation schemas, type guards
- Added to discriminated union, valid event type set, and schema validation map
- Added unit tests for new type guards (72 tests passing)
- Eliminates ~1,776 SSE validation warnings/day

**2. Remove client-side CloudWatch publishing**
- Removed publishSSEMonitorMetrics import and call from assistant-architect-streaming.tsx
- Client-side JS has no AWS credentials — these calls always fail
- Server-side ADOT/CloudWatch metrics remain unaffected
- Eliminates 52 "Credential is missing" warnings

**3. Downgrade ContentSafetyBlockedError to warn level**
- Moved ContentSafetyBlockedError check before generic log.error() in all 4 route handlers
- Content safety blocks are expected behavior, not application errors
- Prevents inflating error counts in CloudWatch dashboards
- Updated: nexus/chat, assistant-architect/execute, nexus/decision-chat, chat-helpers

## Test Plan
- [x] TypeScript typecheck passes (0 errors)
- [x] ESLint passes (0 errors, only pre-existing warnings)
- [x] SSE event type unit tests pass (72/72)
- [ ] Deploy to dev and verify SSE validation warnings drop to near-zero
- [ ] Verify CloudWatch credential warnings stop
- [ ] Verify ContentSafetyBlockedError appears at warn level, not error

Closes #835

PR #859, opened 2026-03-13
fix: remove client-logger import from shared sse-event-types module

krishagel:
## Summary

Fixes production build failure introduced in PR #841 (merged 2026-03-11).

**Error:**
```
Error: Attempted to call createLogger() from the server but createLogger is on the client.
Failed to collect page data for /api/assistant-architect/execute/scheduled
```

**Root cause:** `lib/streaming/sse-event-types.ts` imported `createLogger` from `@/lib/client-logger` (which has `'use client'`). This file is also imported by server-side API routes via `lib/streaming/` imports. Next.js 16 classifies the module as client code due to the transitive `'use client'` dependency, then fails when server routes try to call `createLogger()`.

**Fix:** Removed the `client-logger` import. Replaced the single `log.warn` call with `console.warn`. This file is shared between client and server contexts — it cannot import from either logger.

**Bisection evidence:**
- Build passes at `e05267e6` (before PR #841)
- Build fails at `03cc45e6` (after PR #841)

## Test Plan
- [x] `bun run build` passes (was failing)
- [x] `bun run typecheck` clean
- [x] No functionality lost — the removed log was a single diagnostic `warn` for unrecognized SSE event types, now `console.warn`

### DIFF
--- a/lib/streaming/__tests__/sse-event-types.test.ts
+++ b/lib/streaming/__tests__/sse-event-types.test.ts
@@ -84,7 +84,7 @@ describe('SSE Event Parsing', () => {
       const data = '{"type":"unknown-type","data":"test"}';
 
       // parseSSEEvent should not throw — unknown types are gracefully forwarded
-      // with a structured log.warn via @/lib/client-logger (no console.warn)
+      // with a console.warn (can't use client-logger or server logger in this shared module)
       expect(() => parseSSEEvent(data)).not.toThrow();
       const event = parseSSEEvent(data);
       expect(event.type).toBe('unknown-type');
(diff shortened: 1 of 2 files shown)


krishagel:



### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Must Fix):** 1 found, 1 resolved
- **P3 (Suggestions):** 1 found, 1 resolved

### Changes Made

**P2 - Stale test comment** (TypeScript reviewer)
- `lib/streaming/__tests__/sse-event-types.test.ts:87`: Comment referenced old `client-logger` import. Updated to reflect `console.warn`.

**P3 - Restore structured context in console.warn** (Gemini code suggestion)
- Applied code suggestion: pass `{ type, hint }` object to `console.warn` instead of just the type string.

### Agents Invoked
- Architecture strategist: No blocking issues. Minor SRP note (pre-existing, out of scope).
- Code simplicity reviewer: No issues with PR change. Pre-existing concerns flagged as out of scope.
- TypeScript reviewer: Found stale test comment (fixed). Confirmed fix is correct.

### Not Fixed (Out of Scope)
- `sse-monitoring.ts` and `graceful-degradation.ts` also import `client-logger` but are client-only (verified, no server imports).

### Testing
- 77/77 SSE event type tests pass
- TypeScript: 0 errors
- Build passes

PR #896, opened 2026-04-11
feat: #875 — Voice transcript persistence with Bedrock guardrails

### LINKED ISSUE #875: feat: Voice Transcript Persistence + Bedrock Guardrails Integration
## Summary

Convert voice conversation transcripts into standard Nexus messages, run them through Bedrock content safety guardrails, and save to the existing conversation infrastructure so users can seamlessly continue in text mode. This is issue 4 of 5 in the Nexus Voice Mode epic.

## User Story

As a Nexus user, when I finish a voice conversation, I want the full transcript saved to my conversation so I can review it, search it, and continue the discussion in text mode.

### Transcript Collection

During the voice session, the `GeminiLiveProvider` accumulates transcript entries via `inputAudioTranscription` and `outputAudioTranscription` callbacks:

```typescript
interface TranscriptEntry {
  role: 'user' | 'assistant';
  text: string;
  timestamp: Date;
}
```

When the voice session ends (user clicks "End Conversation" or session times out), the complete ordered transcript is available.

### Transcript → Nexus Messages Conversion

Convert transcript entries into standard `nexus_messages` format:

```typescript
// Each transcript entry becomes a nexus_message
for (const entry of transcript) {
  await saveMessage({
    conversationId: conversationId,
    role: entry.role,
    parts: [{ type: 'text', text: entry.text }],  // Standard MessagePart format
    metadata: { source: 'voice' },  // Optional: mark as voice-originated
    createdAt: entry.timestamp,
  });
}
```

**Key requirements:**
- Messages saved in chronological order with correct timestamps
- Use existing `nexus_messages` table schema (no schema changes needed)
- Parts use the standard `{ type: 'text', text: string }` format
- Conversation `messageCount` and `lastMessageAt` updated after save
- Conversation `modelUsed` updated to reflect the voice model

### Conversation Record Updates

**For existing conversations (voice started mid-thread):**
- Messages append to existing conversation
- Update `messageCount`, `lastMessageAt`, `totalTokens` (if available)

**For new conversations (voice started fresh):**
- Conversation record created at voice session start (issue #874)
- After transcript save, update title (auto-generate from first few exchanges)
- Set `provider` to the voice provider identifier

### Bedrock Guardrails Integration

Run transcript text through existing Bedrock content safety guardrails before saving:

```typescript
// Before saving transcript
const filteredTranscript = await applyGuardrails(transcript);

// applyGuardrails checks each message against Bedrock guardrail
// Uses existing guardrail infrastructure from /lib/guardrails/
// If content blocked → mark that entry, don't save blocked content
// Log blocked content for admin review
```

**Guardrail considerations:**
- Run guardrails on the FULL transcript text before saving (batch check)
- If individual messages are flagged, handle gracefully:
  - Replace flagged content with "[Content filtered by safety policy]"
  - Log the filtering event
  - Don't block the entire transcript — save what's clean
- Use existing `ContentBlockedFetchError` pattern for consistency
- Guardrail settings from existing `Settings.getContentSafety()` 

**Known guardrail constraints (from project learnings):**
- At least one content filter must be non-NONE (HATE at LOW is current minimum)
- PROMPT_ATTACK filter disabled due to 75% false positive rate
- INSULTS/MISCONDUCT at LOW
- All topics set to detect-only

### Seamless Text Continuation

After voice transcript is saved, the user returns to the chat view:
- Conversation loads with all messages (text + voice transcript)
- User can immediately type a new message to continue
- The text AI model sees the full history including voice messages
- No visual distinction needed between voice and text messages (per decision: minimal metadata)

### Transaction Safety

Transcript save must be atomic:
```typescript
await executeTransaction(async (tx) => {
  // Save all transcript messages
  for (const entry of filteredTranscript) {
    await tx.insert(nexusMessages).values({...});
  }
  // Update conversation metadata
  await tx.update(nexusConversations)
    .set({ messageCount: sql`${nexusConversations.messageCount} + ${filteredTranscript.length}`, ... })
    .where(eq(nexusConversations.id, conversationId));
}, "saveVoiceTranscript");
```

### Architecture
- New service: `/lib/voice/transcript-service.ts` — handles conversion, guardrails, persistence
- Reuses existing Drizzle ORM patterns (`executeQuery`, `executeTransaction`)
- Integrates with existing guardrails infrastructure
- Called by the voice session lifecycle (issue #874) on session end

### Performance
- Transcript save should be fast — batch insert, single transaction
- Guardrail check may add 1-3 seconds for the API call — show "Saving conversation..." in UI
- For long voice sessions (100+ exchanges), batch guardrail checks

### Data Integrity
- Use transaction for all-or-nothing save
- Handle edge case: voice session ends unexpectedly (save whatever transcript we have)
- Idempotency: if save is retried, don't duplicate messages (use unique IDs)

## Acceptance Criteria

- [ ] Voice transcript converted to standard nexus_messages format
- [ ] Messages saved with correct chronological ordering and timestamps
- [ ] Conversation metadata (messageCount, lastMessageAt, modelUsed) updated
- [ ] New conversations get auto-generated title from voice content
- [ ] Bedrock guardrails applied to transcript before saving
- [ ] Flagged content replaced with filter notice, not silently dropped
- [ ] Guardrail filtering logged for admin review
- [ ] Transaction ensures atomic save (all messages or none)
- [ ] After save, user can continue conversation in text mode
- [ ] Text AI model sees voice messages in conversation history
- [ ] Partial transcript saved on unexpected disconnection
- [ ] `bun run lint` and `bun run typecheck` pass

## Testing Plan

- **Unit tests:** Transcript → message conversion, guardrail integration (mocked), title generation
- **Integration tests:** Full save flow with test database, transaction rollback on error
- **E2E tests:** Complete voice → save → continue in text flow
- **Manual testing:** Guardrail triggering with test content, unexpected disconnection recovery

## Dependencies

- Issue #872 (Voice Provider Abstraction — provides transcript format)
- Issue #874 (Voice Session Lifecycle — triggers transcript save on session end)
- Existing guardrails: `/lib/guardrails/` and Bedrock guardrail stack
- Existing message save patterns: `/app/api/nexus/messages/`

## References

- nexus_messages schema: `/lib/db/schema/tables/nexus-messages.ts`
- nexus_conversations schema: `/lib/db/schema/tables/nexus-conversations.ts`
- Guardrails stack: `/infra/lib/guardrails-stack.ts`
- ContentBlockedFetchError pattern: `/lib/errors/`
- Drizzle transaction pattern: `/lib/db/drizzle-client.ts`

## Epic

Nexus Voice Mode (Issue 4 of 5)

krishagel:
## Summary

Implements voice transcript persistence (Issue #875, part 4 of the Nexus Voice Mode epic). When a voice session ends, the full transcript is converted to standard Nexus messages, run through Bedrock content safety guardrails, and saved atomically to the database.

- **Transcript Service** (`lib/voice/transcript-service.ts`): Prepares entries (filter non-final, merge consecutive same-role), applies guardrails per-entry with graceful degradation, saves all messages + conversation metadata in a single transaction
- **WS Handler Integration**: Captures transcript from provider state before disconnect, resolves userId, triggers async save during cleanup (non-blocking, non-fatal)
- **Guardrails**: Blocked content replaced with "[Content filtered by safety policy]" — entries are never silently dropped. User entries checked via `checkInputSafety`, assistant entries via `checkOutputSafety`
- **Title Generation**: Auto-generates from first (non-filtered) user message for new conversations

## Changes
- `lib/voice/transcript-service.ts` — New: core persistence service
- `lib/voice/__tests__/transcript-service.test.ts` — New: 19 unit tests
- `lib/voice/ws-handler.ts` — Modified: integrated transcript save into session lifecycle
- `lib/voice/index.ts` — Modified: added exports

## Test Plan
- [x] 19 unit tests pass (entry preparation, guardrail integration, title generation, error handling, graceful degradation)
- [x] `bun run typecheck` passes (zero errors)
- [x] `bun run lint` passes (zero errors, existing warnings only)
- [ ] Manual testing: complete voice session → verify messages appear in conversation
- [ ] Manual testing: verify text continuation after voice transcript save
- [ ] Manual testing: verify guardrail filtering with test content

Closes #875

### DIFF
--- a/lib/constants/conversation.ts
+++ b/lib/constants/conversation.ts
@@ -1,0 +1,16 @@
+/**
+ * Conversation Constants
+ *
+ * Shared sentinel values used across conversation creation, title detection,
+ * and voice transcript persistence. Centralizing these prevents silent
+ * coupling failures if a default value changes in one location.
+ */
+
+/**
+ * Default title assigned to new conversations when no title is provided.
+ *
+ * Used by:
+ * - nexus-conversations.ts (conversation creation default)
+ * - transcript-service.ts (detect untitled conversations for auto-titling)
+ */
+export const DEFAULT_CONVERSATION_TITLE = "New Conversation"
--- a/lib/db/drizzle/nexus-conversations.ts
+++ b/lib/db/drizzle/nexus-conversations.ts
@@ -25,6 +25,7 @@ import {
   nexusConversationEvents,
 } from "@/lib/db/schema";
 import { countAsInt } from "@/lib/db/drizzle/helpers/pagination";
+import { DEFAULT_CONVERSATION_TITLE } from "@/lib/constants/conversation";
 import type {
   NexusConversationMetadata,
   NexusFolderSettings,
@@ -280,7 +281,7 @@ export async function createConversation(data: CreateConversationData) {
   }
 
   // Use provided title or default (handles undefined, null, empty, or whitespace-only strings)
-  const title = data.title?.trim() || "New Conversation";
+  const title = data.title?.trim() || DEFAULT_CONVERSATION_TITLE;
 
   const result = await executeQuery(
     (db) =>
--- a/lib/voice/__tests__/transcript-service.test.ts
+++ b/lib/voice/__tests__/transcript-service.test.ts
@@ -1,0 +1,437 @@
+/**
+ * Tests for Voice Transcript Persistence Service
+ *
+ * Tests transcript preparation (merge, filter), guardrail integration (mocked),
+ * title generation, and the full save flow with mocked DB operations.
+ *
+ * Issue #875
+ */
+
+import { saveVoiceTranscript, prepareTranscriptEntries } from "../transcript-service"
+import type { TranscriptEntry } from "../types"
+
+// ============================================
+// Mocks
+// ============================================
+
+const mockExecuteTransaction = jest.fn()
+const mockGetConversationById = jest.fn()
+
+jest.mock("@/lib/db/drizzle-client", () => ({
+  executeTransaction: (...args: unknown[]) => mockExecuteTransaction(...args),
+}))
+
+jest.mock("@/lib/db/drizzle/nexus-conversations", () => ({
+  getConversationById: (...args: unknown[]) => mockGetConversationById(...args),
+}))
+
+jest.mock("@/lib/db/schema", () => ({
+  nexusMessages: { id: "id", conversationId: "conversation_id", role: "role" },
+  nexusConversations: { id: "id", messageCount: "message_count" },
+}))
+
+jest.mock("@/lib/db/json-utils", () => ({
+  safeJsonbStringify: (val: unknown) => JSON.stringify(val),
+}))
+
+// Mock drizzle-orm's sql template tag and eq function
+jest.mock("drizzle-orm", () => ({
+  sql: Object.assign(
+    (strings: TemplateStringsArray, ...values: unknown[]) => ({ strings, values, _tag: "sql" }),
+    { raw: (s: string) => ({ raw: s, _tag: "sql" }) },
+  ),
+  eq: (a: unknown, b: unknown) => ({ a, b, _tag: "eq" }),
+}))
+
+const mockCheckInputSafety = jest.fn()
+const mockCheckOutputSafety = jest.fn()
+const mockIsGuardrailsEnabled = jest.fn()
+
+jest.mock("@/lib/safety", () => ({
+  getContentSafetyService: () => ({
+    isGuardrailsEnabled: mockIsGuardrailsEnabled,
+    checkInputSafety: mockCheckInputSafety,
+    checkOutputSafety: mockCheckOutputSafety,
+  }),
+}))
+
+jest.mock("@/lib/logger", () => ({
+  createLogger: () => ({
+    info: jest.fn(),
+    warn: jest.fn(),
+    error: jest.fn(),
+    debug: jest.fn(),
+  }),
+  generateRequestId: () => "test-request-id",
+  startTimer: () => jest.fn().mockReturnValue(42),
+}))
+
+// ============================================
+// Test Helpers
+// ============================================
+
+function makeEntry(
+  role: "user" | "assistant",
+  text: string,
+  opts?: { isFinal?: boolean; timestamp?: Date },
+): TranscriptEntry {
+  return {
+    role,
+    text,
+    isFinal: opts?.isFinal ?? true,
+    timestamp: opts?.timestamp ?? new Date("2026-04-11T10:00:00Z"),
+  }
+}
+
+// ============================================
+// Tests: prepareTranscriptEntries
+// ============================================
+
+describe("prepareTranscriptEntries", () => {
+  it("should filter out non-final entries", () => {
+    const entries = [
+      makeEntry("user", "hello", { isFinal: false }),
+      makeEntry("user", "hello world", { isFinal: true }),
+      makeEntry("assistant", "partial", { isFinal: false }),
+      makeEntry("assistant", "hi there", { isFinal: true }),
+    ]
+
+    const result = prepareTranscriptEntries(entries)
+    expect(result).toHaveLength(2)
+    expect(result[0].text).toBe("hello world")
+    expect(result[1].text).toBe("hi there")
+  })
+
+  it("should filter out empty text entries", () => {
+    const entries = [
+      makeEntry("user", "", { isFinal: true }),
+      makeEntry("user", "  ", { isFinal: true }),
+      makeEntry("user", "actual content", { isFinal: true }),
+    ]
+
+    const result = prepareTranscriptEntries(entries)
+    expect(result).toHaveLength(1)
+    expect(result[0].text).toBe("actual content")
+  })
+
+  it("should merge consecutive same-role entries", () => {
+    const t1 = new Date("2026-04-11T10:00:00Z")
+    const t2 = new Date("2026-04-11T10:00:01Z")
+    const t3 = new Date("2026-04-11T10:00:02Z")
+
+    const entries = [
+      makeEntry("user", "hello", { timestamp: t1 }),
+      makeEntry("user", "how are you?", { timestamp: t2 }),
+      makeEntry("assistant", "I'm good", { timestamp: t3 }),
+    ]
+
+    const result = prepareTranscriptEntries(entries)
+    expect(result).toHaveLength(2)
+    expect(result[0].role).toBe("user")
+    expect(result[0].text).toBe("hello how are you?")
+    // Keeps the earlier timestamp
+    expect(result[0].timestamp).toBe(t1)
+    expect(result[1].role).toBe("assistant")
+    expect(result[1].text).toBe("I'm good")
+  })
+
+  it("should not merge entries with different roles", () => {
+    const entries = [
+      makeEntry("user", "hello"),
+      makeEntry("assistant", "hi"),
+      makeEntry("user", "another question"),
+    ]
+
+    const result = prepareTranscriptEntries(entries)
+    expect(result).toHaveLength(3)
+  })
+
+  it("should return empty array for empty input", () => {
+    expect(prepareTranscriptEntries([])).toHaveLength(0)
+  })
+
+  it("should return empty array when all entries are non-final", () => {
+    const entries = [
+      makeEntry("user", "partial", { isFinal: false }),
+      makeEntry("assistant", "also partial", { isFinal: false }),
+    ]
+    expect(prepareTranscriptEntries(entries)).toHaveLength(0)
+  })
+
+  it("should cap at 500 entries", () => {
+    const entries: TranscriptEntry[] = []
+    for (let i = 0; i < 600; i++) {
+      entries.push(
+        makeEntry(i % 2 === 0 ? "user" : "assistant", `message ${i}`),
+      )
+    }
+
+    const result = prepareTranscriptEntries(entries)
+    expect(result.length).toBeLessThanOrEqual(500)
+  })
+
+  it("should trim whitespace from entry text", () => {
+    const entries = [
+      makeEntry("user", "  hello world  "),
+    ]
+
+    const result = prepareTranscriptEntries(entries)
+    expect(result[0].text).toBe("hello world")
+  })
+})
+
+// ============================================
+// Tests: saveVoiceTranscript
+// ============================================
+
+describe("saveVoiceTranscript", () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockIsGuardrailsEnabled.mockReturnValue(false)
+    mockCheckInputSafety.mockResolvedValue({ allowed: true, processedContent: "test" })
+    mockCheckOutputSafety.mockResolvedValue({ allowed: true, processedContent: "test" })
+  })
+
+  it("should throw if conversation not found", async () => {
+    mockGetConversationById.mockResolvedValue(null)
+
+    const entries = [makeEntry("user", "hello")]
+    await expect(
+      saveVoiceTranscript("conv-123", 1, entries),
+    ).rejects.toThrow("Conversation conv-123 not found")
+  })
+
+  it("should return early with zero counts for empty transcript", async () => {
+    mockGetConversationById.mockResolvedValue({ id: "conv-123", title: "Test" })
+
+    const result = await saveVoiceTranscript("conv-123", 1, [])
+    expect(result.messageCount).toBe(0)
+    expect(result.filteredCount).toBe(0)
+    expect(result.titleGenerated).toBe(false)
+    expect(mockExecuteTransaction).not.toHaveBeenCalled()
+  })
+
+  it("should return early for transcript with only non-final entries", async () => {
+    mockGetConversationById.mockResolvedValue({ id: "conv-123", title: "Test" })
+
+    const entries = [
+      makeEntry("user", "partial", { isFinal: false }),
+    ]
+    const result = await saveVoiceTranscript("conv-123", 1, entries)
+    expect(result.messageCount).toBe(0)
+    expect(mockExecuteTransaction).not.toHaveBeenCalled()
+  })
+
+  it("should save transcript without guardrails when disabled", async () => {
+    mockGetConversationById.mockResolvedValue({ id: "conv-123", title: "Existing Title" })
+    mockIsGuardrailsEnabled.mockReturnValue(false)
+    mockExecuteTransaction.mockImplementation(async (fn: (tx: unknown) => Promise<void>) => {
+      const mockTx = {
+        insert: jest.fn().mockReturnValue({ values: jest.fn() }),
+        update: jest.fn().mockReturnValue({ set: jest.fn().mockReturnValue({ where: jest.fn() }) }),
+      }
+      await fn(mockTx)
+    })
+
+    const entries = [
+      makeEntry("user", "hello"),
+      makeEntry("assistant", "hi there"),
+    ]
+
+    const result = await saveVoiceTranscript("conv-123", 1, entries, "gemini-2.0-flash-live-001")
+    expect(result.messageCount).toBe(2)
+    expect(result.filteredCount).toBe(0)
+    expect(result.titleGenerated).toBe(false)
+    expect(mockExecuteTransaction).toHaveBeenCalledTimes(1)
+    // Guardrail checks should not have been called
+    expect(mockCheckInputSafety).not.toHaveBeenCalled()
+    expect(mockCheckOutputSafety).not.toHaveBeenCalled()
+  })
+
+  it("should generate title for new conversations", async () => {
+    mockGetConversationById.mockResolvedValue({ id: "conv-123", title: "New Conversation" })
+    mockIsGuardrailsEnabled.mockReturnValue(false)
+
+    let capturedTx: Record<string, jest.Mock> | null = null
+    mockExecuteTransaction.mockImplementation(async (fn: (tx: unknown) => Promise<void>) => {
+      capturedTx = {
+        insert: jest.fn().mockReturnValue({ values: jest.fn() }),
+        update: jest.fn().mockReturnValue({
+          set: jest.fn().mockReturnValue({ where: jest.fn() }),
+        }),
+      }
+      await fn(capturedTx)
+    })
+
+    const entries = [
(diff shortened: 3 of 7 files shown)


krishagel:

### Summary
- Critical Issues: 0
- High Priority: 2
- Low Priority: 3

---

None identified.

---

**H1. Missing NaN guard on `Number.parseInt` for userId**
- **File**: `lib/voice/ws-handler.ts:166`
- **Problem**: `getUserIdByCognitoSub` returns `string | null`. If the DB returns a non-numeric string (data corruption, unexpected value), `Number.parseInt("abc", 10)` returns `NaN`. This `NaN` userId is then passed to `getConversationById(conversationId, NaN)` where the Drizzle `eq(nexusConversations.userId, NaN)` comparison will never match any row (SQL: `WHERE user_id = NaN` is always false), so the save silently fails with "conversation not found". While not exploitable, it masks a real error.
- **Fix**: Add a `Number.isNaN` check after parsing and return `null` if the parsed value is not a valid positive integer.

**H2. Information leakage in error message — userId exposed in thrown error**
- **File**: `lib/voice/transcript-service.ts:114`
- **Problem**: The error message `Conversation ${conversationId} not found or not owned by user ${userId}` includes the numeric userId. This error is caught in `ws-handler.ts` cleanup and logged via `error.message`. While the log is server-side (not sent to client), the error propagates to the `.catch()` handler which logs it. If this error were ever surfaced to a client (through a future refactor), it would leak internal user IDs and conversation UUIDs. The `conversationId` is less concerning since the caller already knows it, but including `userId` is unnecessary.
- **Fix**: Remove userId from the error message. Use `Conversation not found` or log userId separately in structured log context.

---

**L1. `voiceModel` parameter hardcoded to `"voice-model"` / `"gemini-live"` in guardrail call**
- **File**: `lib/voice/transcript-service.ts:303`
- **Problem**: `checkOutputSafety` receives hardcoded strings `"voice-model"` and `"gemini-live"` instead of the actual model/provider values. The `voiceModel` parameter (e.g., `"gemini-2.0-flash-live-001"`) is available on the function but not threaded down to `applyGuardrails`. This means guardrail audit logs won't reflect the actual model that generated the content.
- **Fix**: Pass the `voiceModel` parameter through to `applyGuardrails` and use it in the `checkOutputSafety` call.

**L2. Unused `GUARDRAIL_BATCH_SIZE` doesn't actually rate-limit — batches run concurrently within each batch, but batches are sequential**
- **File**: `lib/voice/transcript-service.ts:66, 291-293`
- **Problem**: The batching pattern uses `Promise.all` inside each batch (up to 20 concurrent guardrail API calls) but processes batches sequentially. For a 500-entry transcript, this means 25 sequential batches of 20 concurrent calls. This is functional but the batch size constant name is misleading — it's a concurrency limit, not a batch size. Minor naming/documentation concern only.
- **Fix**: Rename constant or add a comment clarifying it controls concurrency, not payload batching.

**L3. `cleanup()` captures transcript reference before disconnect but uses fire-and-forget pattern — no guarantee of completion**
- **File**: `lib/voice/ws-handler.ts:414-437`
- **Problem**: The `saveVoiceTranscript` call is fire-and-forget (`.then()/.catch()` with no await). If the Node.js process is shutting down (e.g., ECS task draining, container replacement), the save may be interrupted mid-transaction. The transaction itself is atomic (Drizzle `executeTransaction`), so partial writes are not a concern — but the entire save could be lost. This is documented as intentional ("non-blocking, non-fatal") which is reasonable for the current use case.
- **Fix**: Consider logging a warning if the save takes longer than a threshold (e.g., 5s) to help identify slow persistence under load. No code change required — this is an observability suggestion.

---

### Data Integrity & Compliance Scan

**PII Exposure**: No PII is logged in the transcript service. User content is processed through guardrails before persistence. The `sanitizeForLogging` utility is used where appropriate in `ws-handler.ts`.

**FERPA Compliance**: Voice transcripts are saved with user ownership verification (`getConversationById` checks both `conversationId` and `userId`). Users can only access their own conversations. Content safety filtering adds a layer of protection for student-generated content.

**Access Control**: Properly enforced — authentication via session cookie, authorization via `hasToolAccess("voice-mode")`, ownership via `getConversationById(id, userId)`.

---

- **Ownership verification**: `getConversationById` checks both conversation existence AND user ownership before allowing transcript save
- **Atomic transaction**: All messages + metadata saved in a single `executeTransaction` — no partial state on failure
- **Graceful degradation**: Guardrail failures allow content through rather than blocking the save entirely (consistent with existing patterns)
- **Content safety**: Blocked content is replaced with a placeholder, never silently dropped — maintains conversation flow integrity
- **No client exposure**: Error details stay server-side; client receives only generic "Voice provider error" messages
- **Idempotent cleanup**: `sessionEnded` flag prevents double-save of transcripts
- **Proper logging**: Uses `@/lib/logger` throughout (no `console.*` calls)
- **Input validation**: UUID format check, conversation ID length limit, audio data size validation all remain intact from prior PRs
- **Object.create(null)**: Cookie parsing uses prototype-free object (existing code, not new, but good to note)

---

### Required Actions
1. Fix all high priority issues (H1, H2)
2. Fix all low priority issues (L1, L2, L3)
3. Run security checks: `bun run lint`, `bun run typecheck`
4. Verify all tests pass after fixes

---
*Automated security review by security-analyst-specialist agent*

krishagel:



### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Must Fix):** 7 found → 7 resolved
- **P3 (Suggestions):** 4 found → 4 fixed

### Changes Made

**P2 fixes:**
1. Replaced manual `parseInt` + NaN guard with canonical `getUserIdByCognitoSubAsNumber` utility (ws-handler.ts)
2. Threaded `voiceModel` AND `voiceProvider` through `saveVoiceTranscript` → `applyGuardrails` → `checkOutputSafety` — no more hardcoded `"voice-model"` / `"gemini-live"`
3. Replaced hardcoded `"New Conversation"` with `DEFAULT_CONVERSATION_TITLE` constant
4. Replaced `Record<string, unknown>` updateFields with inline spread object for Drizzle type safety
5. Removed userId/conversationId from thrown error message — now generic "Conversation not found or access denied"
6. Shallow copy transcript array before `provider.disconnect()` to prevent data loss
7. Registered WS message handler BEFORE `sendToClient(ws, { type: "ready" })` and before `resolveTranscriptContext` async DB lookup — fixes race condition for early audio frames

**P3 fixes:**
1. Capture actual elapsed time on early-return path instead of hardcoding `processingTimeMs: 0`
2. Removed redundant `if (messageValues.length > 0)` guard
3. Added truncation warning log when transcript entries exceed MAX_TRANSCRIPT_ENTRIES
4. Added null check for `voiceSettings.model` before passing to `resolveTranscriptContext`

### Review Agents Used
- Security analyst (PII/information leakage, ownership verification, access control)
- Architecture review (type safety, established patterns, canonical utilities)
- Code quality (redundant guards, misleading values, shared constants)

### Testing
- All 113 voice tests passing (21 transcript-service, 26 ws-handler, 66 others)
- 2 new tests: voiceModel/voiceProvider pass-through verification + fallback string verification
- Lint: 0 errors (pre-existing warnings only)
- Typecheck: clean

Ready for re-review.

krishagel:



**Incremental review** — only processed new feedback since Round 1 (2026-04-11T17:49:28Z)

### Severity Summary
- **P1 (Blocking):** 0 found
- **P3 (Suggestions):** 4 found → 4 fixed

### Changes Made

**P2 fixes:**
1. **processedContent now used from guardrail results** — When `checkInputSafety`/`checkOutputSafety` returns transformed content (e.g., PII redaction), the processed text is now persisted instead of the original. Currently a no-op (quick check methods don't transform), but defensive against future safety service changes.
2. **`DEFAULT_CONVERSATION_TITLE` extracted to shared constant** — New `lib/constants/conversation.ts` imported by both `nexus-conversations.ts` (creation default) and `transcript-service.ts` (title detection). Eliminates silent coupling.
3. **`voiceProvider` added to initial save log** — Both `voiceModel` and `voiceProvider` now logged at transcript save start for consistent audit trail.

**P3 fixes:**
1. **Tests use shared constant** — All test assertions reference `DEFAULT_CONVERSATION_TITLE` instead of hardcoded `"New Conversation"`.
2. **Truncation warning test added** — Verifies `prepareTranscriptEntries` emits `warn()` with `droppedCount` when entries exceed MAX_TRANSCRIPT_ENTRIES.
3. **Short-session race documented** — Added code comment in ws-handler.ts explaining the known limitation where cleanup before `resolveTranscriptContext` completes skips persistence.
4. **Logger mock refactored** — Exposed mock logger functions for assertion using `__mockLogger` pattern to avoid Jest hoisting TDZ issues.

### Review Agents Used
- Architecture review (shared constants, type safety)
- Code quality (processedContent gap, observability, test coverage)

### Testing
- All 115 voice tests passing (23 transcript-service, 26 ws-handler, 66 others)
- 2 new tests: processedContent pass-through + truncation warning log
- Lint: 0 errors (pre-existing warnings only)
- Typecheck: clean

Ready for re-review.

krishagel:



**Incremental review** — only processed new feedback since Round 2 (2026-04-11T22:29:11Z)

### Severity Summary
- **P1 (Blocking):** 0 found
- **P3 (Suggestions):** 5 found → 5 fixed

### Changes Made

**P2 fixes:**
1. **Race condition: transcriptContext timing window closed** — Store the `resolveTranscriptContext` promise on session state and await it in `cleanupSession`. Early client disconnects no longer silently lose transcripts — cleanup waits for the in-flight DB lookup before deciding whether to persist.
2. **Guardrail timeout added** — Wrapped all guardrail processing in a 30s `Promise.race` timeout. On timeout, falls back to saving unfiltered content rather than hanging during Bedrock brownouts. Prevents the server from blocking indefinitely with 500 entries x 25 batches.
3. **processingTimeMs bug fixed** — `startTimer()` returns `void`, not a number. The `typeof elapsed === "number"` guard was always false, so `processingTimeMs` was always `0`. Replaced with manual `Date.now() - startTime` delta.

**P3 fixes:**
1. **providerRef JSDoc** — Added comment explaining why the ref-wrapper pattern is used (message handler needs to read the provider after post-connect assignment).
2. **getContentSafetyService singleton comment** — Added inline comment noting it's a cached singleton, no heavy init on repeated calls.
3. **Title test comment fixed** — Changed misleading "Title should be 40 chars + ..." to accurate "at most 40 chars + ...; trim() may make it shorter".
4. **applyGuardrailsBatched parameter refactor** — Grouped options into object to satisfy max-params lint rule (was 6 params, now 3).
5. **3 new tests added** — See below.

### New Tests
- **DB error propagation**: `getConversationById` throwing (connection error) vs returning null
- **processedContent fallback**: Verifies `??` operator when `processedContent` is null/undefined
- **Batch boundary crossing**: 25 entries across `GUARDRAIL_CONCURRENCY_LIMIT` of 20 — no entries dropped

### Review Agents Used
- Code quality (processingTimeMs bug, dead code, observability)
- Architecture review (race condition, timeout resilience, parameter structure)

### Testing
- All 118 voice tests passing (26 transcript-service, 26 ws-handler, 66 others)
- 3 new tests: DB error propagation + processedContent fallback + batch boundary
- Lint: 0 errors (334 pre-existing warnings)
- Typecheck: clean

Ready for re-review.

krishagel:



**Incremental review** — only processed new feedback since Round 3 (2026-04-11T22:43:55Z)

### Severity Summary

### Changes Made:

**P1 — Ping interval leak after early disconnect (ws-handler.ts)**
Added `session.sessionEnded` guard before starting the ping interval after `await session.transcriptContextPromise`. If the WebSocket closed during that await, cleanup would have already run and cleared the interval — but the new interval started unconditionally with no way to clear it.

**P1 — Unhandled rejection from async cleanup (ws-handler.ts)**
Attached `.catch()` to `cleanup()` calls in both `ws.on("close")` and `ws.on("error")` handlers. Since `cleanupSession` is async, the returned Promise was being silently dropped, risking unhandled rejections on unexpected throws.

**P2 — setTimeout leak in Promise.race (transcript-service.ts)**
Added `try/finally` with `clearTimeout` around the `Promise.race` in `applyGuardrails`. When `applyGuardrailsBatched` resolves before the 30s timeout, the timer reference was lost and the callback would fire on a settled Promise — a harmless but unnecessary resource leak.

**P3 — Tail capping for MAX_TRANSCRIPT_ENTRIES (transcript-service.ts)**
Changed `merged.slice(0, MAX_TRANSCRIPT_ENTRIES)` to `merged.slice(-MAX_TRANSCRIPT_ENTRIES)` so long sessions preserve the most recent entries rather than the earliest. Updated test to verify tail entries are kept.

**P3 — transcriptContextPromise comment clarification (ws-handler.ts)**
Expanded the comment at the `transcriptContextPromise` assignment to explain why the promise is stored separately even though it's also immediately awaited — needed for the cleanup fallback path on early disconnects.

### Not addressed:
- **E2E test gap** — acknowledged as a valid gap per CLAUDE.md testing requirements. This requires a running voice session with a live provider, which is beyond unit test scope. Recommend creating a separate issue for E2E voice test infrastructure.

### Testing:
- All 26 unit tests passing
- Lint: 0 errors (335 pre-existing warnings, no new warnings)
- TypeScript: clean

Ready for re-review.

krishagel:
## Review Feedback Addressed (Round 5)



**Incremental review** — only processed new feedback since Round 4 (2026-04-11T22:52:29Z)

### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Must Fix):** 3 found -> 3 resolved
- **P3 (Suggestions):** 4 found -> 4 fixed

### Changes Made

**P2 fixes:**
1. **Early exit in cleanupSession** — Added guard before the `transcriptContextPromise` await: if there's no context, no in-flight promise, and no captured transcript, return immediately. Prevents unnecessary DB round-trips for sessions that closed during setup (auth failure, config timeout, etc.).
2. **`guardrailsBypassed` field added to `TranscriptSaveResult`** — Callers can now audit/alert when content was saved without safety filtering (disabled, timed out, or errored). The field is set to `true` when guardrails are disabled or when the entire pipeline times out/fails.
3. **Guardrail timeout/error log elevated to error level** — Previously `warn`, now `error` since content is being persisted without safety filtering. Includes `conversationId` for audit trail.

**P3 fixes:**
1. **`prepareTranscriptEntries` removed from index.ts export** — Tests import directly from `./transcript-service`; no external consumers needed it on the public API surface.
2. **`requestId` added to truncation warning** — `prepareTranscriptEntries` now accepts optional `requestId` parameter, included in the truncation log for request correlation in log aggregation.
3. **`modelUsed` column verified + 2 tests added** — Confirmed `nexusConversations.modelUsed` exists in schema (`model_used varchar(100)`). Tests verify the column is written when `voiceModel` is provided and omitted when it's not.
4. **Cross-reference comment on sessionEnded ping guard** — Added note linking behavioral tests to E2E coverage tracking in issue #875.

### Not Addressed
- **E2E test gap** — Acknowledged across all review rounds. Requires live voice provider infrastructure. Already tracked in issue #875.

### Review Agents Used
- Code quality (guardrail bypass surfacing, export reduction, log context)
- Schema verification (modelUsed column validation)

### Testing
- All 120 voice tests passing (28 transcript-service, 26 ws-handler, 66 others)
- 2 new tests: modelUsed column write verification (present + absent)
- guardrailsBypassed assertions added to 4 existing tests
- Lint: 0 errors (335 pre-existing warnings, no new warnings)
- Typecheck: clean

Ready for re-review.

krishagel:
## Review Feedback Addressed (Round 6)



**Incremental review** — only processed new feedback since Round 5 (2026-04-12T03:44:19Z)

### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Must Fix):** 6 found -> 6 resolved
- **P3 (Suggestions):** 4 found -> 4 fixed

### Changes Made

**P2 fixes:**
1. **resolveTranscriptContext timeout added (ws-handler.ts)** — Wrapped the `getUserIdByCognitoSubAsNumber` DB lookup in a 5s `Promise.race` timeout (`TRANSCRIPT_CONTEXT_TIMEOUT_MS`). Prevents the in-flight promise from blocking cleanup resource release if the DB is slow. On timeout, transcript persistence is silently skipped (non-fatal).
2. **Title truncation off-by-ellipsis fixed (transcript-service.ts)** — `MAX_TITLE_LENGTH` is 40 total INCLUDING the "..." suffix. Previously `slice(0, 40)` + "..." could produce 43-char titles. Now slices to `MAX_TITLE_LENGTH - 3` before appending ellipsis.
3. **processedContent null/undefined warn log added** — When the safety service returns `processedContent == null`, a warn-level log is emitted distinguishing "intentionally null" from "API shape changed". The original text is still used as fallback.
4. **100% filtered entries elevated to error log** — When all entries in the transcript are filtered by guardrails (`allFiltered: true`), the log is elevated from warn to error level. Operators can detect potential guardrail misconfiguration or systematic abuse.
5. **Conversation-not-found warn log added** — Added explicit warn log at the throw site when `getConversationById` returns null, distinguishing "conversation deleted mid-session" from generic errors.
6. **Double-await race safety documented** — Expanded comment in `cleanupSession` explaining why the `sessionEnded` flag prevents state corruption when both `handleVoiceConnection` and `cleanupSession` await the same `transcriptContextPromise`.

**P3 fixes:**
1. **Transaction rollback test** — New test verifying that when the transaction callback throws (e.g., unique constraint violation), the error propagates to the caller.
2. **All-entries-filtered test** — New test verifying that 100% filtered entries trigger error-level log with `allFiltered: true`.
3. **processedContent null warn test** — New test verifying that null/undefined `processedContent` from safety service triggers warn log.
4. **Title truncation test updated** — Assertion changed from `<= 43` to `<= 40` to match the corrected behavior.

### Not Addressed
- **E2E test gap** — Acknowledged across all review rounds. Requires live voice provider infrastructure. Already tracked in issue #875.
- **Guardrail bypass alerting beyond logging** — The `guardrailsBypassed` field and error-level logging are in place. External alerting (CloudWatch alarms) is an ops concern outside this PR's scope.
- **Concurrent save deduplication** — Would require schema changes (unique constraint on conversation + voice session). Low risk in practice since `sessionEnded` flag prevents double-save within a single session.

### Review Agents Used
- Code quality (timeout patterns, log levels, edge case detection)
- Architecture review (race condition documentation, title truncation contract)

### Testing
- All 123 voice tests passing (31 transcript-service, 26 ws-handler, 66 others)
- 3 new tests: transaction rollback + all-entries-filtered + processedContent null warn
- Lint: 0 errors (335 pre-existing warnings, no new warnings)
- Typecheck: clean

Ready for re-review.

PR #932, opened 2026-04-30
fix(agent): stream long-running invocations and serialize concurrent turns

krishagel:
## Summary
- AgentCore container now streams its response (async-generator entrypoint with periodic heartbeats) so morning briefs and other multi-minute invocations no longer hit the ~5-min idle ceiling on buffered synchronous responses. Headroom is now the documented 15-min `InvokeAgentRuntime` quota, capped at 13 min client-side.
- Both Lambdas (`agent-cron` and `agent-router`) now consume the SSE stream, discard heartbeat events, and keep the final `result` event. Falls back to JSON parsing if the container ever serves a buffered response. Cron Lambda also surfaces the real error class/message in the chat fallback string and logs response Content-Type so future regressions are visible.
- New `psd-agent-session-locks-${env}` DynamoDB table + per-session lock in the router. Two messages back-to-back from the same user/space serialize on `sessionId` (1s poll, up to 13 min), so the second message no longer collides with the first on AgentCore's sticky-routed microVM and surface as `"I processed your message but had no response."`. TTL backstop frees a wedged lock if a Lambda dies mid-turn.
- Restores CloudWatch logging on the cron Lambda role (Apr 24 regression — inline IAM policy granted logs perms on `/aws/lambda/psd-agent-cron:*` instead of `/aws/lambda/psd-agent-cron-${env}:*`, silently denying CreateLogStream/PutLogEvents). Attaches the AWS-managed `AWSLambdaBasicExecutionRole` to bypass the broken inline rule without renaming the role.

## Why
Morning brief was failing every day with `"Agent temporarily unavailable for scheduled task."` Aurora `agent_scheduled_runs` showed every failed run hitting exactly ~302,000 ms (5:01) — a hard infrastructure idle timeout, not the documented 15-min API quota. The Apr 28 brief succeeded at 4:09; the new `hagelk-morning-brief` skill (3 Chat spaces + Gmail + Calendar + news) tipped it past the 5-min ceiling. Separately, sending two chat messages back-to-back was returning `"I processed your message but had no response."` for the second message — root-caused to AgentCore session-ID sticky routing colliding with the OpenClaw turn loop.

## Test plan
- [x] `bunx tsc --noEmit` clean on the infra package
- [x] `bun run lint` — 0 errors / 0 warnings on touched files
- [x] `uv run -m py_compile agentcore_wrapper.py` clean
- [x] `bunx cdk synth AIStudio-AgentPlatformStack-Dev` succeeds
- [x] Image built and pushed: `psd-agent-base-dev:2026-04-30-initial` (digest `sha256:bf0eaf32e6fbb5074a5317d7c661dcfd7479b9b3993fd7aa0211430ad2c01bc7`)
- [x] Deployed to dev; live router logs show `"AgentCore SSE stream complete"` with `haveResult=true` on real chat traffic
- [x] Orphaned duplicate EventBridge schedule (`f2d8a3bd-...`) deleted; only `fb0e8fb4-...` remains
- [ ] Tomorrow's 6:00 AM PT scheduled morning brief delivers full content (not the fallback) and `agent_scheduled_runs` records `status='success'` with `latency_ms` between 4–13 min
- [ ] Send two chat messages back-to-back: both receive distinct, real responses; router logs show second invocation acquiring the session lock after the first releases
- [ ] CloudWatch produces fresh log streams under `/aws/lambda/psd-agent-cron-dev` on the next scheduled fire

## Deploy
```
cd infra && bunx cdk deploy AIStudio-AgentPlatformStack-Dev \
  --context agentImageTag=2026-04-30-initial \
  --context agentImageDigest=sha256:bf0eaf32e6fbb5074a5317d7c661dcfd7479b9b3993fd7aa0211430ad2c01bc7
```

### DIFF
--- a/infra/agent-image/agentcore_wrapper.py
+++ b/infra/agent-image/agentcore_wrapper.py
@@ -256,6 +256,20 @@ def main():
         """
         Handle an agent invocation from AgentCore.
 
+        Yields events as a streaming response (text/event-stream). Async-
+        generator entrypoints make BedrockAgentCoreApp emit SSE so bytes flow
+        on the wire continuously — required for invocations that exceed the
+        ~5-minute idle ceiling on buffered synchronous responses (the morning
+        brief takes 4–8 minutes when it scans Chat / Gmail / Calendar).
+
+        Stream contract:
+          - first yield: {"type": "start"} — flushes headers immediately
+          - heartbeats:  {"type": "heartbeat", "elapsed_s": int} every ~30s
+          - final yield: {"result": "...", "metadata": {...}}
+
+        The cron Lambda discards heartbeat events and uses the final event's
+        `result` field. See infra/lambdas/agent-cron/index.ts.
+
         Expected payload keys (all optional except `prompt`):
             prompt                    — the user's text
             user_email                — caller's email (used as stable identity)
@@ -286,7 +300,8 @@ def main():
         )
 
         if not user_message.strip():
-            return {"result": "I didn't receive a message. Could you try again?"}
+            yield {"result": "I didn't receive a message. Could you try again?"}
+            return
 
         # First invocation for a new workspace prefix → pull memory from S3.
         if workspace_prefix and workspace_prefix != _current_workspace_prefix:
@@ -355,17 +370,34 @@ def main():
         else:
             framed = f"{now_header}\n\n{user_message}"
 
+        # Flush the SSE headers immediately. Without an early yield the SDK
+        # waits for the first chunk before sending headers, defeating the
+        # streaming purpose.
+        invocation_start = time.time()
+        yield {"type": "start", "session_id": session_id}
+
         loop = asyncio.get_running_loop()
-        result = await loop.run_in_executor(
+        process_task = loop.run_in_executor(
             None, adapter.process, framed, session_id, model_override
         )
 
+        # Heartbeat every 30s while adapter.process runs in the executor.
+        # Each yield writes bytes to the SSE response, keeping the connection
+        # alive past any infrastructure idle timeout.
+        while True:
+            try:
+                result = await asyncio.wait_for(asyncio.shield(process_task), timeout=30)
+                break
+            except asyncio.TimeoutError:
+                elapsed = int(time.time() - invocation_start)
+                yield {"type": "heartbeat", "elapsed_s": elapsed}
+
         logger.info(
-            "Invocation complete: session=%s response_length=%d",
-            session_id, len(result),
+            "Invocation complete: session=%s response_length=%d elapsed_s=%d",
+            session_id, len(result), int(time.time() - invocation_start),
         )
 
-        return {
+        yield {
             "result": result,
             "metadata": {
                 "session_id": session_id,
--- a/infra/agent-image/agentcore_wrapper.py
+++ b/infra/agent-image/agentcore_wrapper.py
@@ -267,8 +267,9 @@ def main():
           - heartbeats:  {"type": "heartbeat", "elapsed_s": int} every ~30s
           - final yield: {"result": "...", "metadata": {...}}
 
-        The cron Lambda discards heartbeat events and uses the final event's
-        `result` field. See infra/lambdas/agent-cron/index.ts.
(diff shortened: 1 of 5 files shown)


krishagel:



### Severity Summary
- **P2 (Must Fix):** 6 found → 6 resolved

### Changes Made

**P1 — Lock ownership race condition** (flagged by Gemini, Codex, Copilot, and Claude):
- Added a random `lockToken` (UUID) to each DynamoDB PutItem on lock acquisition
- `releaseSessionLock` now uses a conditional delete (`ConditionExpression` on `lockToken`) so only the current owner can release — prevents stale holders from deleting a newer holder's lock after TTL expiry
- Changed `waitForSessionLock` return type from `boolean` to `string|null` (token on success, null on timeout)
- Both call sites (owner path and cross-user path) now check the return value and send a user-friendly "busy" message instead of proceeding without the lock
- Added exponential backoff to lock polling (1s → 2s → 4s → 8s cap) reducing DDB read volume ~75% for long waits

**P2 — SSE parser hardening:**
- Added `TextDecoder.flush()` call after the for-await loop in both `consumeAgentCoreStream` implementations
- Normalized `\r\n` and `\r` to `\n` before splitting SSE events (spec compliance)
- Explicitly handle the `start` event type instead of silent fallthrough
- Replaced `buffer.trim().length > 0` with `buffer.length > 0` to avoid unnecessary allocation
- Added cross-reference comments between the two copies explaining the duplication is intentional (separate build pipelines)

**P2 — Error message sanitization:**
- Cron Lambda error response no longer leaks internal error class names and stack traces to user DMs

**P2 — IAM least-privilege:**
- Replaced the overly broad `AWSLambdaBasicExecutionRole` managed policy with a narrow inline policy targeting only the actual log group

### Review Agents Used:
- Manual review of all 4 automated reviewers (Gemini, Codex, Copilot, Claude x2)
- Synthesized overlapping feedback into unified fixes

### Testing:
- TypeScript compilation passes for both agent-router and agent-cron
- CDK synth passes for all stacks
- Full project lint (0 errors) and typecheck passing
- No new warnings introduced in modified files

Ready for re-review. Thank you for the feedback!

krishagel:



**Incremental review** — only processed new feedback since Round 1 (2026-05-01T20:23:20Z)

### Severity Summary
- **P1 (Blocking):** 0 found → 0 resolved (none new)
- **P3 (Suggestions):** 3 found → 3 fixed

### Changes Made

**P2 — Stale JSDoc blocks before `tryAcquireSessionLock`:**
- Removed orphaned JSDoc belonging to `isDuplicateMessage` that was displaced during Round 1 refactoring
- Removed obsolete JSDoc describing the old boolean lock interface
- Restored proper JSDoc on `isDuplicateMessage` function

**P2 — Sentinel strings as lock tokens are fragile:**
- Introduced `LOCK_PASS_THROUGH` named constant replacing magic strings `'no-table'` and `'ddb-error'`
- `releaseSessionLock` now explicitly short-circuits on `LOCK_PASS_THROUGH` instead of relying on coincidental `ConditionalCheckFailedException` behavior when the sentinel never matched a real DDB row

**P2 — `agentResult` implicitly `any` in lock paths:**
- Both call sites (owner + cross-user) now use `const agentResult = await invokeAgentCore(...).finally(() => releaseSessionLock(...))` — properly typed, no intermediate `let`

**P3 — Lock TTL best-effort documentation:**
- Added JSDoc documenting that serialization is best-effort for turns exceeding 14 min (TTL expiry allows re-entry in the 14-15 min window)

**P3 — Improved SYNC comments on `consumeAgentCoreStream`:**
- Both copies now reference the specific function name (`consumeAgentCoreStream`)
- Document the known intentional differences between the two copies (Response vs {body: unknown}, elapsed-time logging)

**P3 — DDB poll volume (item #6):** No action — acknowledged as informational and acceptable at current scale

### Review Agents Used:
- Manual review of Claude reviewer feedback (post-Round-1 comment)

### Testing:
- Lint: 0 errors (332 pre-existing warnings, none new)
- TypeScript typecheck: passing
- CDK synth: passing for all stacks

Ready for re-review. Thank you for the feedback!

PR #1004, opened 2026-05-28
feat(agent-platform): deep telemetry, dashboard tabs, email triage

krishagel:
## Summary
- Adds **Triage** and **Conversations** tabs to /admin/agents.
- Fixes the empty/wrong existing tabs: Usage (real model + tokens + latency), Cost (tag-activation guidance), Patterns (truthful empty states), Credentials (paste-secret modal).
- Captures **full conversation transcripts + tool-call timelines** for every agent turn (new schema + writes from the harness through the router).
- **Auto-registers bundled image skills** so the Skills tab lists chat-card / psd-github / psd-workspace / etc. instead of only greetings-demo.
- Lands the **email triage Phase 1** feature (consolidates work from prior sessions that was untracked on `dev`).
- Adds a 90-day **retention sweep Lambda** for the deep-telemetry tables.

## Telemetry pipeline — top to bottom

```
OpenClaw events → harness_adapter.TurnResult → agentcore_wrapper metadata
   → agent-router invokeAgentCore() → logTelemetry()
     ├─ agent_messages (model, tokens_in, tokens_out, latency_ms)
     ├─ agent_message_content (one row per role/content, 64KB cap)
     └─ agent_tool_invocations (one row per tool call, args+result jsonb)
```

The harness extracts model id from agent event payloads, sums token usage from whichever event carries it, measures latency from chat.send → final, and builds a per-turn tool timeline from `event:agent` `stream="tool_call"` / `stream="tool_result"` events. The wrapper passes everything through to the router with a legacy-string fallback so older harness builds still write usable rows.

## What's in each tab now
| Tab | Status |
|---|---|
| Usage | Real model / tokens / latency once new image deploys. |
| Cost | Honest empty-state with one-click link to activate ` costCenter` in Billing (must run from payer account, AWS Organizations limitation). |
| Patterns | Three differentiated empty-states. |
| Credentials | Paste-secret modal — provisions Secrets Manager + flips status + audit log, atomically. |
| Skills | Auto-registers every bundled image skill on each deploy. |
| **Triage (new)** | Lists opted-in users + per-user Manage link. |
| **Conversations (new)** | Full transcript + tool timeline drilldown, filterable. |

## Schema
- Migration **078**: ` agent_message_content` + ` agent_tool_invocations`. CASCADE on ` agent_messages`; indexes on session_id / user_email / tool_name + created_at.
- Daily prune Lambda (` psd-agent-telemetry-prune-<env>`) — 04:00 UTC, 90-day retention, 5000-row batches. Aggregated ` agent_messages` summary rows kept indefinitely.

## Email triage (Phase 1)
- DynamoDB ` psd-agent-triage-<env>` (per-user state).
- ` psd-agent-triage-poll-<env>` Lambda — 5-min schedule, rules + Nova Micro classifier, label apply + Chat escalation cards, ` @psd/Task` gesture → AgentCore handoff (90s timeout, per-message claim).
- ` psd-agent-triage-digest-<env>` Lambda — per-user daily Scheduler.
- ` psd-email-triage` skill — enable/disable/configure from Chat.
- ` /admin/agents/[userEmail]/triage` — admin detail with Pause / Reset / Re-onboard.

## Cost-allocation tag (one-time manual step, payer-account only)
` ` ` 
aws ce update-cost-allocation-tags-status \
  --cost-allocation-tags-status TagKey=costCenter,Status=Active
` ` ` 

## Test plan
After deploy (canonical command, current image tag, then rebuild + redeploy the image once for the new harness):

- [ ] Migration 078 runs; both new tables exist.
- [ ] Skill initializer Custom Resource invokes the Lambda; ` psd_agent_skills` populates with every bundled skill (scope=shared, scan_status=clean).
- [ ] Triage tab lists opted-in users; Manage link opens the existing per-user detail page.
- [ ] Cost tab shows the activation explanation + Billing console link.
- [ ] Patterns tab shows the right differentiated empty-state.
- [ ] Credentials tab → click Provision on a pending request → modal → paste secret + pick scope → secret lands in Secrets Manager + request flips fulfilled + audit row.
- [ ] After agent-image rebuild + redeploy: send an agent message → Usage tab shows real model id (e.g. ` anthropic.claude-3-5-sonnet-...`) with non-zero tokens + correct latency. Conversations tab → find the session → drilldown shows transcript + tool call timeline.
- [ ] ` psd-agent-telemetry-prune-<env>` exists with the daily 04:00 UTC EventBridge schedule. Invoke with ` RETENTION_DAYS=0` to verify cleanup path.

### DIFF
--- a/actions/admin/agent-conversations.actions.ts
+++ b/actions/admin/agent-conversations.actions.ts
@@ -120,8 +120,10 @@ export async function listAgentConversations(
     )
 
     // Tool-call counts in one extra query, then merge.
+    // Use Object.create(null) to avoid prototype pollution from DB-sourced
+    // sessionId keys (see docs/guides/silent-failure-patterns.md).
     const sessionIds = rows.map((r) => r.sessionId)
-    let toolCounts: Record<string, number> = {}
+    const toolCounts: Record<string, number> = Object.create(null)
     if (sessionIds.length > 0) {
       const toolRows = await executeQuery(
         (db) =>
@@ -135,7 +137,7 @@ export async function listAgentConversations(
             .groupBy(agentToolInvocations.sessionId),
         "agentConversations.toolCounts",
       )
-      toolCounts = Object.fromEntries(toolRows.map((r) => [r.sessionId, r.count]))
+      for (const r of toolRows) toolCounts[r.sessionId] = r.count
     }
 
     const items: ConversationListItem[] = rows.map((r) => ({
--- a/actions/admin/agent-conversations.actions.ts
+++ b/actions/admin/agent-conversations.actions.ts
@@ -34,7 +34,7 @@ export interface ConversationListItem {
   totalInputTokens: number
   totalOutputTokens: number
   models: string[]
-  hasError: boolean
+  hasGuardrailBlock: boolean
 }
 
 export interface ConversationDetailMessage {
@@ -109,7 +109,7 @@ export async function listAgentConversations(
             totalInputTokens: sql<number>`COALESCE(SUM(${agentMessages.inputTokens}), 0)::int`,
             totalOutputTokens: sql<number>`COALESCE(SUM(${agentMessages.outputTokens}), 0)::int`,
             models: sql<string[]>`array_remove(array_agg(DISTINCT ${agentMessages.model}), NULL)`,
-            hasError: sql<boolean>`bool_or(${agentMessages.guardrailBlocked})`,
+            hasGuardrailBlock: sql<boolean>`bool_or(${agentMessages.guardrailBlocked})`,
           })
           .from(agentMessages)
           .where(conditions)
@@ -155,7 +155,7 @@ export async function listAgentConversations(
       totalInputTokens: r.totalInputTokens,
       totalOutputTokens: r.totalOutputTokens,
       models: r.models ?? [],
-      hasError: r.hasError ?? false,
+      hasGuardrailBlock: r.hasGuardrailBlock ?? false,
     }))
 
     timer({ status: "success" })
@@ -204,6 +204,10 @@ export async function getAgentConversationDetail(
 
     // Content + tool queries are independent — run in parallel to cut
     // latency by ~2x (both keyed by sessionId, no dependency).
+    // Limits prevent an extremely long session (500+ turns at 64KB/row)
+    // from pulling 32MB+ into ECS memory.
+    const CONTENT_LIMIT = 500
+    const TOOL_LIMIT = 1000
     const [contentRows, toolRows] = await Promise.all([
       executeQuery(
         (db) =>
@@ -211,7 +215,8 @@ export async function getAgentConversationDetail(
             .select()
             .from(agentMessageContent)
             .where(eq(agentMessageContent.sessionId, sessionId))
-            .orderBy(agentMessageContent.createdAt),
+            .orderBy(agentMessageContent.createdAt)
+            .limit(CONTENT_LIMIT),
         "agentConversations.detail.content",
       ),
       executeQuery(
@@ -220,7 +225,8 @@ export async function getAgentConversationDetail(
             .select()
             .from(agentToolInvocations)
             .where(eq(agentToolInvocations.sessionId, sessionId))
-            .orderBy(agentToolInvocations.startedAt),
+            .orderBy(agentToolInvocations.startedAt)
+            .limit(TOOL_LIMIT),
         "agentConversations.detail.tools",
       ),
     ])
--- a/actions/admin/agent-conversations.actions.ts
+++ b/actions/admin/agent-conversations.actions.ts
@@ -1,0 +1,277 @@
+"use server"
+
+/**
+ * Admin server actions for the Conversations tab on /admin/agents.
+ *
+ * Reads from agent_messages (summary) + agent_message_content (full text)
+ * + agent_tool_invocations (timeline). Admin-only. Capped queries to
+ * keep the dashboard snappy; deep filtering can be added later when we
+ * see what admins actually search for.
+ *
+ * Privacy contract: viewing requires the `administrator` role. The tables
+ * have a 90-day retention sweep (see agent-telemetry-prune Lambda) so
+ * the blast radius is bounded.
+ */
+
+import { and, desc, eq, gte, sql } from "drizzle-orm"
+
+import { requireRole } from "@/lib/auth/role-helpers"
+import { handleError, createSuccess } from "@/lib/error-utils"
+import { createLogger, generateRequestId, startTimer } from "@/lib/logger"
+import { executeQuery } from "@/lib/db/drizzle-client"
+import { agentMessages } from "@/lib/db/schema/tables/agent-messages"
+import { agentMessageContent } from "@/lib/db/schema/tables/agent-message-content"
+import { agentToolInvocations } from "@/lib/db/schema/tables/agent-tool-invocations"
+import type { ActionState } from "@/types"
+
+export interface ConversationListItem {
+  sessionId: string
+  userId: string
+  startedAt: string
+  lastTurnAt: string
+  turnCount: number
+  toolCallCount: number
+  totalInputTokens: number
+  totalOutputTokens: number
+  models: string[]
+  hasError: boolean
+}
+
+export interface ConversationDetailMessage {
+  role: string
+  contentText: string
+  contentTruncated: boolean
+  createdAt: string
+}
+
+export interface ConversationDetailTool {
+  toolName: string
+  status: string
+  errorText: string | null
+  durationMs: number
+  startedAt: string
+  finishedAt: string
+  toolArgs: Record<string, unknown> | null
+  toolResult: Record<string, unknown> | null
+}
+
+export interface ConversationDetail {
+  sessionId: string
+  userId: string
+  turns: Array<{
+    messageId: number
+    model: string | null
+    createdAt: string
+    latencyMs: number
+    inputTokens: number
+    outputTokens: number
+    messages: ConversationDetailMessage[]
+    tools: ConversationDetailTool[]
+  }>
+}
+
+const DEFAULT_DAYS = 7
+
+/**
+ * List sessions, newest activity first. Each row is one session with
+ * aggregated counters across all of its turns. Defaults to the last 7
+ * days so the table stays fast; admin can scope wider via filters.
+ */
+export async function listAgentConversations(
+  daysBack: number = DEFAULT_DAYS,
+  userIdFilter?: string,
+): Promise<ActionState<ConversationListItem[]>> {
+  const requestId = generateRequestId()
+  const timer = startTimer("listAgentConversations")
+  const log = createLogger({ requestId, action: "listAgentConversations" })
+
+  try {
+    await requireRole("administrator")
+    const days = Math.min(Math.max(1, daysBack), 90)
+    const cutoff = new Date(Date.now() - days * 86400_000)
+
+    const conditions = userIdFilter
+      ? and(
+          gte(agentMessages.createdAt, cutoff),
+          eq(agentMessages.userId, userIdFilter.toLowerCase()),
+        )
+      : gte(agentMessages.createdAt, cutoff)
+
+    const rows = await executeQuery(
+      (db) =>
+        db
+          .select({
+            sessionId: agentMessages.sessionId,
+            userId: agentMessages.userId,
+            startedAt: sql<string>`MIN(${agentMessages.createdAt})::text`,
+            lastTurnAt: sql<string>`MAX(${agentMessages.createdAt})::text`,
+            turnCount: sql<number>`COUNT(*)::int`,
+            totalInputTokens: sql<number>`COALESCE(SUM(${agentMessages.inputTokens}), 0)::int`,
+            totalOutputTokens: sql<number>`COALESCE(SUM(${agentMessages.outputTokens}), 0)::int`,
+            models: sql<string[]>`array_remove(array_agg(DISTINCT ${agentMessages.model}), NULL)`,
+            hasError: sql<boolean>`bool_or(${agentMessages.guardrailBlocked})`,
+          })
+          .from(agentMessages)
+          .where(conditions)
+          .groupBy(agentMessages.sessionId, agentMessages.userId)
+          .orderBy(desc(sql`MAX(${agentMessages.createdAt})`))
+          .limit(200),
+      "agentConversations.list",
+    )
+
+    // Tool-call counts in one extra query, then merge.
+    const sessionIds = rows.map((r) => r.sessionId)
+    let toolCounts: Record<string, number> = {}
+    if (sessionIds.length > 0) {
+      const toolRows = await executeQuery(
+        (db) =>
+          db
+            .select({
+              sessionId: agentToolInvocations.sessionId,
+              count: sql<number>`COUNT(*)::int`,
+            })
+            .from(agentToolInvocations)
+            .where(gte(agentToolInvocations.startedAt, cutoff))
+            .groupBy(agentToolInvocations.sessionId),
+        "agentConversations.toolCounts",
+      )
+      toolCounts = Object.fromEntries(toolRows.map((r) => [r.sessionId, r.count]))
+    }
+
+    const items: ConversationListItem[] = rows.map((r) => ({
+      sessionId: r.sessionId,
+      userId: r.userId,
+      startedAt: r.startedAt,
+      lastTurnAt: r.lastTurnAt,
+      turnCount: r.turnCount,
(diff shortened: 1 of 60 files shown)


krishagel:

### Summary
- **Critical Issues:** 1
- **High Priority:** 3
- **Low Priority:** 3

---

#### C1: `toolCounts` accumulator uses `Object.fromEntries` on user-controlled keys
**File:** `actions/admin/agent-conversations.actions.ts:138`
**Problem:** `toolCounts` is built via `Object.fromEntries(toolRows.map(...))` where `sessionId` values come from the database. Per the project's silent-failure-patterns guide: "Don't use `{}` as an accumulator for model/user-controlled keys — use `Object.create(null)`". While the data comes from an admin-only query and the risk is low in practice, the pattern is explicitly called out in the codebase rules.
**Fix:** Replace `Object.fromEntries(toolRows.map(...))` with an accumulator built on `Object.create(null)`:
```typescript
const toolCounts: Record<string, number> = Object.create(null)
for (const r of toolRows) toolCounts[r.sessionId] = r.count
```
**Reference:** `docs/guides/silent-failure-patterns.md`

---

#### H1: Conversation detail renders raw `contentText` without sanitization
**File:** `app/(protected)/admin/agents/_components/agent-conversations.tsx:304-309`
**Problem:** The `ConversationDetailView` renders `m.contentText` (agent/user message content from the database) inside a `<pre>` tag with no sanitization. While `<pre>` with `whitespace-pre-wrap` and React's JSX escaping provides basic XSS protection (React auto-escapes text children), the content may include `toolArgs` and `toolResult` rendered via `JSON.stringify` in the same component (lines 344-356). If any tool result contains HTML entities or Unicode control characters that interact poorly with the rendering context, it could cause display corruption. The `dangerouslySetInnerHTML` is not used, so actual XSS risk is low with React, but the component should still use the project's `text-sanitizer.ts` utility for defense-in-depth.
**Fix:** Import and apply `sanitizeText` from `@/lib/utils/text-sanitizer` to `contentText` before rendering. Same for `errorText` in tool call details.
**Reference:** OWASP XSS Prevention, project `lib/utils/text-sanitizer.ts`

#### H2: Triage poll Lambda uses full DynamoDB Scan for user profile lookup
**File:** `infra/lambdas/agent-triage-poll/storage.ts:111-137`
**Problem:** `getUserProfile()` performs a full table Scan with a `FilterExpression` to find a user by email in the USERS_TABLE. At scale (hundreds of users, thousands of DDB items), this is both expensive (consumes read capacity across the entire table every call) and slow. The function is called once per task gesture, and the 5-minute profile cache mitigates repeated hits, but the underlying pattern is a known DynamoDB anti-pattern. The agent-router Lambda uses an `email-index` GSI on the same table (`QueryCommand` with `IndexName: 'email-index'`), proving the GSI exists.
**Fix:** Replace the Scan with a Query against the `email-index` GSI, matching the pattern in `agent-router/index.ts:882-903` (`resolveUserByEmailPrefix`). This is O(1) instead of O(N).
**Reference:** AWS DynamoDB best practices — avoid Scan when an index exists

#### H3: Permission boundary allows `Resource: "*"` for Bedrock actions in region-restriction exception
**File:** `infra/lib/constructs/security/permission-boundaries/dev-boundary.json:133-136` and `prod-boundary.json`
**Problem:** The PR adds `bedrock:InvokeModel`, `bedrock:InvokeModelWithResponseStream`, `bedrock:Converse`, and `bedrock:ConverseStream` to the `NotAction` list in the `DenyRegionRestriction` statement. This means these Bedrock actions are now allowed in **any region** without restriction. While this was intentionally done to support cross-region inference profiles (the triage Lambda's Nova Micro calls were routing through us-west-2), it also means any Lambda or role under this boundary can invoke any Bedrock model in any region. The triage Lambda's own IAM policy correctly scopes to specific models (`nova-micro`, `haiku`), so the real blast radius is bounded by the resource-level policies. However, the boundary is shared across all roles in the environment — any future Lambda that gets `bedrock:InvokeModel` in its resource policy could invoke models in unexpected regions.
**Fix:** Consider adding a comment to the boundary file documenting that Bedrock region restrictions are enforced at the resource-policy level (per-Lambda), not at the boundary level, and why. No code change needed if this is an accepted architectural decision.

---

#### L1: Multiple `@ts-expect-error` suppressions in triage actions
**File:** `actions/admin/agent-triage.actions.ts:123, 212, 297, 332, 372`
**Problem:** Five separate `@ts-expect-error` comments suppress a nested `@smithy/types` version mismatch between `@aws-sdk/lib-dynamodb` and `@aws-sdk/client-dynamodb`. While the comments explain the reason, version mismatches between AWS SDK packages can cause subtle runtime bugs (marshalling differences, missing fields). The same pattern is used successfully elsewhere in the codebase, but it would be cleaner to resolve the version mismatch.
**Fix:** Check if upgrading `@aws-sdk/lib-dynamodb` and `@aws-sdk/client-dynamodb` to the same minor version resolves the type conflict. If not, the current suppression is acceptable but should be tracked as tech debt.

#### L2: Telemetry prune Lambda SQL injection surface via `sql()` identifier
**File:** `infra/lambdas/agent-telemetry-prune/index.ts:87-88`
**Problem:** The `pruneTable` function uses `${sql(table)}` to interpolate the table name, which is the correct postgres.js pattern for identifiers (not `sql.unsafe()`). The `table` parameter is constrained to a union type `"agent_message_content" | "agent_tool_invocations"`, making injection impossible. No fix needed — this is a positive practice observation. Noted for completeness since the function accepts a string parameter.

#### L3: `claimTaskGesture` recursive retry without depth limit
**File:** `infra/lambdas/agent-triage-poll/storage.ts:195-200`
**Problem:** When the `taskGestureClaims` map doesn't exist yet on the DynamoDB item, the function initializes it and then recursively calls itself. If the initialization succeeds but the retry still hits the same `ValidationException` (e.g., a concurrent writer), this could loop. In practice the condition is extremely unlikely (it requires the map to be deleted between the init and the retry), and the DynamoDB conditional expression would break the cycle, but a depth guard would be safer.
**Fix:** Add a `depth` parameter (default 0) and bail after depth >= 1.

---

- **Admin role enforcement**: All server actions in `agent-conversations.actions.ts`, `agent-credentials.actions.ts`, and `agent-triage.actions.ts` correctly call `requireRole("administrator")` before any data access.
- **Timing-safe auth**: The feedback API route (`/api/agent/feedback/route.ts`) uses `timingSafeEqual` for bearer token comparison, preventing timing attacks.
- **Input validation**: Credential name regex (`CREDENTIAL_NAME_RE`), secret value length limits, status filter allowlisting in credentials actions.
- **Audit trail**: Every credential provisioning action writes to `psd_agent_credentials_audit` with the admin's identity resolved from session.
- **Data retention**: 90-day prune Lambda with batch-delete pattern prevents unbounded table growth. Content is capped at 64KB per message.
- **Session isolation**: Cross-user agent invocations hash the invoker's email into the session ID, preventing information leakage between different users consulting the same agent.
- **Parameterized queries**: All Drizzle ORM and postgres.js queries use parameterized values (no string interpolation in SQL).
- **Secret value never logged**: The credential provisioning flow explicitly avoids logging the secret value — only metadata (secret path, action type) is logged.
- **Privacy contract**: Signal table writes for the Organizational Nervous System explicitly exclude user identifiers and message content, only storing aggregated {building, topic, count}.
- **Concurrency control**: Triage poll Lambda has `reservedConcurrentExecutions: 1` preventing duplicate processing from concurrent invocations.

### Required Actions
1. Fix critical issue C1 (Object.create(null) for accumulator)
2. Fix high priority H1 (sanitize conversation content rendering)
3. Fix high priority H2 (replace Scan with GSI Query in storage.ts)
4. Address H3 with documentation if accepted as architectural decision
5. Run security checks: `bun run lint`, `bun run typecheck`
6. Verify all tests pass after fixes

---
*Automated security review by security-analyst-specialist agent*

krishagel:



### Severity Summary
- **P1 (Blocking):** 7 found → 7 resolved
- **P2 (Must Fix):** 12 found → 12 resolved
- **P3 (Suggestions):** 6 found → 6 fixed

### Changes Made (12 files, 232 additions, 108 deletions):

**P1 — Critical:**
- Scoped toolCounts query to returned sessionIds via `inArray()` — eliminates unbounded table scan
- Separated `userFilter` from `activeUserFilter` to prevent keystroke-triggered server queries
- Added `void` before async `load()` in useEffect for no-floating-promises compliance
- Atomic `WHERE status='pending'` + `executeTransaction` on credential provisioning — prevents double-provisioning race
- Added `SAFE_PATH_SEGMENT_RE` validation on `requestedBy` — prevents Secrets Manager path traversal
- Aligned Secrets Manager tags with IAM allowlist (Environment, ManagedBy, OwnerEmail)
- Fixed `createdAt` being overwritten on feedback upsert (removed `sql` import)

**P2 — Must Fix:**
- Added `startTimer()` to all 5 triage server actions (was imported but never called)
- Added `sanitizeForLogging()` to all userEmail log calls per PII contract
- Added `ConditionExpression: attribute_exists(userEmail)` to prevent ghost DynamoDB rows
- Fixed `detectCorrection` to use `direction` parameter — inbox-removal corrections were silently dropped
- Added rolling trim (max 20) to `recentTaskCreations` — was growing unbounded
- Lazy-resolve `dmSpaceName` in escalation path — new users' escalations were silently skipped
- Added missing tags to `AgentTelemetryPruneSchedule` EventBridge rule
- Fixed `pruneLambdaRole` to use `vpcEnabled: true` instead of manual policy append
- Fixed misleading batch size comment + removed hardcoded local paths

**P3 — Low Priority:**
- Stable React keys instead of array indices for messages/tools
- `sanitizeForLogging` on sessionId logging
- Parallelized content+tool DB queries via `Promise.all`
- Added `message_id` indexes for CASCADE delete performance
- Removed local filesystem path from docs
- Added `void` before `refresh()` in triage detail callback

### Review Agents Used:
- All feedback categories addressed manually (security, performance, correctness, architecture)
- Sources: gemini-code-assist, Copilot, Codex, CodeQL, Claude

### Testing:
- All 47 triage-poll Lambda tests passing
- Lint: 0 errors (347 pre-existing warnings, down from 348)
- TypeScript: 0 errors
- CI will validate on push

Ready for re-review. Thank you for the thorough feedback!

krishagel:



**Incremental review** — only processed new feedback since Round 1 (2026-05-28T13:57:07Z)

### Severity Summary
- **P1 (Blocking):** 0 found
- **P2 (Must Fix):** 2 found → 2 resolved
- **P3 (Suggestions):** 1 found → 1 fixed

### Changes Made (3 files, 13 additions, 7 deletions):

**P2 — Must Fix:**
- Added `void` before all 3 fire-and-forget `loadAll()` calls in `credentials-client.tsx` — prevents unhandled promise rejections and matches codebase convention
- Used collision-proof React keys `${turn.messageId}-msg-${idx}` and `${turn.messageId}-tool-${idx}` instead of timestamp-based keys that could collide

**P3 — Suggestions:**
- Documented partial consistency window in `provisionCredentialFromRequest` catch block — retry safely hits Secrets Manager rotation path

### Items Reviewed but Not Changed:
- **`@ts-expect-error` for DynamoDB** (5 occurrences): Root cause is `@aws-sdk/client-cost-explorer@3.1035.0` pulling `@smithy/types@4.14.1` while DynamoDB packages at `3.982.0` expect `@smithy/types@4.12.0`. Adding overrides could have cascading effects — the suppressions are pragmatic and the runtime behavior is correct.
- **Migration 078 registration**: Verified present in `infra/database/migrations.json` ✅
- **DynamoDB Scan in `getTriageSummaryList`**: Accepted for Phase 1 per reviewer — MAX_PAGES=5 cap is safety valve
- **`tabLoading` for triage**: Already wired through `buildLoaders` mechanism — works correctly

### Testing:
- Lint: 0 errors (347 pre-existing warnings)
- TypeScript: 0 errors
- All changes verified against current codebase

Ready for re-review.

krishagel:



**Incremental review** — only processed new feedback since Round 2 (2026-05-28T14:06:14Z)

### Severity Summary
- **P2 (Must Fix):** 7 found → 7 resolved

### Changes Made (11 files, 80 additions, 28 deletions):

**P1 — Critical:**
- Capped deep telemetry INSERT fan-out to MAX_MESSAGES_PER_TURN=200 + MAX_TOOLS_PER_TURN=200 in `logTelemetry()` — prevents connection pool exhaustion from misbehaving harness
- Separated labelsAdded/labelsRemoved iteration in `detectCorrection` call site — eliminates O(N²) fragile direction detection via combined-array `.includes()`

**P2 — Must Fix:**
- Added `.limit(500)` / `.limit(1000)` to content/tool queries in `getAgentConversationDetail` — prevents unbounded 32MB+ result sets
- Renamed `hasError` → `hasGuardrailBlock` (actions + UI) — guardrail blocks are not errors
- Added SAFETY_CAP warning log to telemetry prune Lambda — ops can now detect when daily prune falls behind
- Fixed skill-initializer Lambda to use `vpcEnabled: true` instead of manual `addManagedPolicy` workaround
- Added `message_id` FK indexes to Drizzle schema — eliminates schema drift vs. migration 078
- Added `ProjectionExpression` to triage `ScanCommand` — reduces DDB read cost
- Removed unnecessary `as never` casts + stale `@ts-expect-error` directives from chat.ts and digest Lambda — types are actually compatible

**P3 — Low Priority:**
- Added proper `ScheduledHandler` parameter typing to telemetry prune Lambda
- Reduced triage-poll Lambda timeout from 5min to 4.5min (270s) — 30s safety margin prevents silently dropped invocations

### Items Reviewed but Not Changed:
- **`RULES_CONFIDENCE_FLOOR` (N2 nit)**: Reviewer flagged as dead constant, but it IS used at lines 398/402/406 in `classifyAndLabel()` as the LLM confidence threshold
- **`toast` in useEffect deps (N3 nit)**: `toast` is in `useCallback` deps (not directly in useEffect), required by React's exhaustive-deps rule — removing it would be incorrect
- **`@ts-expect-error` for DynamoDB (5 occurrences)**: Unchanged from Round 2 — root cause is cross-package `@smithy/types` version mismatch; suppressions are pragmatic and runtime behavior is correct

### Review Agents Used:
- All feedback categories addressed manually (correctness, performance, architecture, security)
- Sources: 2 new Claude review comments since Round 2

### Testing:
- Lint: 0 errors (347 pre-existing warnings)
- TypeScript: 0 errors
- 47/47 triage-poll Lambda tests passing
- All 3 standalone Lambda typechecks clean (poll, digest, prune)

Ready for re-review. Thank you for the thorough feedback!

krishagel:



**Incremental review** — only processed new feedback since Round 3 (2026-05-28T15:04:59Z)

### Severity Summary
- **P1 (Blocking):** 0 found (NULL::jsonb was a false positive)

### Changes Made (6 files, 49 additions, 22 deletions):

**P2 — Must Fix:**
- Strengthened `userId` type validation in feedback API — `typeof body.userId !== "string"` rejects non-string values that would silently pass the previous falsy check
- Added partial-consistency monitoring to credential provisioning — `smWriteSucceeded` flag + structured `PARTIAL_CONSISTENCY` log.error enables CloudWatch Alarms for the SM-succeeds-but-DB-fails edge case
- Extracted Lambda function names as constants in CDK stack — `cronFunctionName`, `triageDigestFunctionName`, `skillBuilderFunctionName` defined once and referenced by both env vars and Lambda definitions, preventing silent breakage on name changes

**P3 — Low Priority:**
- Added defensive null guard on `finishedAt.toISOString()` in conversation detail — column is NOT NULL but `?.toISOString()` prevents runtime errors from stale data
- Documented `user_email` vs `user_id` naming discrepancy in Drizzle schemas

### Items Reviewed but Not Changed:
- **NULL::jsonb cast** (reviewer flagged as critical): False positive — `NULL::jsonb` is valid PostgreSQL syntax, produces a typed NULL. No runtime error occurs.
- **`@ts-expect-error` for DynamoDB** (5 occurrences): Unchanged from Rounds 2-3 — root cause is cross-package `@smithy/types` version mismatch; suppressions are pragmatic
- **DynamoDB Scan in `listEnabledUsers`**: Already uses GSI (`QueryCommand` with `IndexName: 'email-index'`) with Scan fallback
- **`digestEnabled` defaults to true**: Intentional per SKILL.md design
- **Sequential INSERT in skill-initializer**: Acceptable for ~10 skills
- **`console.error` in llm.ts**: Already uses structured JSON wrapper with `eslint-disable` comment

### Review Agents Used:
- All feedback categories addressed manually (correctness, security, infrastructure)
- Source: 1 new Claude review comment since Round 3

### Testing:
- Lint: 0 errors (347 pre-existing warnings)
- TypeScript: 0 errors
- 47/47 triage-poll Lambda tests passing
- CDK synth error is pre-existing (unrelated VPC policy validation)

Ready for re-review. Thank you for the thorough feedback!

PR #1087, opened 2026-06-30
fix: #1083 - GLM-5 agent usage & cost telemetry (real tokens, model id, token×pricing cost)

### LINKED ISSUE #1083: fix(agent-platform): GLM-5 usage & cost never tracked on /admin/agents — capture tokens, real model id, and token×pricing cost
## Summary

The admin **Agents** dashboard (`/admin/agents`) reports **0 tokens, a single mislabeled `"default"` model row, and ~$0 cost** for the GLM-5–powered agents. The breakage is entirely on the **write side** of telemetry — the dashboard read code is mostly fine; it is being fed empty/garbage data.

As a result we cannot answer the two questions the dashboard exists to answer:
1. **How much usage is GLM-5 actually getting?**
2. **What would it cost if we switched the agents to a different model?**

This follows up the assumptions baked into #890, which assumed "the Router Lambda already captures token usage & model" and "native AWS tags handle cost." Both are false for the GLM-5 / Bedrock Mantle path.

> Verified via direct code reads. The actual model in use is `zai.glm-5` ("GLM-5 (via Bedrock Mantle)"), configured in `infra/agent-image/openclaw.json:11,26`.

## Root causes (all verified in code)

**1. Tokens are always 0.**
- `infra/agent-image/agentcore_wrapper.py:368-375` returns response `metadata` with **no** `input_tokens` / `output_tokens` keys.
- `infra/agent-image/harness_adapter.py` `OpenClawAdapter.process()` is typed `-> str` and returns only the assistant **text** — any usage from the model is discarded.
- `infra/lambdas/agent-router/index.ts:974-975` reads the missing fields as `(metadata.input_tokens as number) || 0` → every `agent_messages` row is written with `input_tokens = 0`, `output_tokens = 0`.

**2. Model is recorded as the literal `"default"`, never GLM-5.**
- Wrapper sets `"model": model_override or "default"` (`agentcore_wrapper.py:373`), and `model_override = payload.get("model")` is always `None` because the router never sends a `model` field (`index.ts:896-909`).
- The router's fallback `model: (metadata.model as string) || 'kimi-k2.5'` (`index.ts:976`) is stale dead code (`"default"` is truthy, so the fallback never fires).
- Net: `getAgentModelBreakdown` (`actions/admin/agent-telemetry.actions.ts:367`) groups everything under one `"default"` bucket with 0 tokens.

**3. No token→cost path, and no GLM-5 pricing row.**
- `agent_messages` (`lib/db/schema/tables/agent-messages.ts`) has `model`, `input_tokens`, `output_tokens` but **no cost column**; `agent-telemetry.actions.ts` does no pricing math and never joins `ai_models`.
- No `ai_models` row exists for `zai.glm-5`; `openclaw.json:17` hardcodes its cost to all zeros.
- So a "cost if we switch models" projection is impossible today.

**4. AWS Cost Explorer can't attribute GLM-5 spend.**
- `actions/admin/agent-cost.actions.ts:62-67` filters Cost Explorer by tag `costCenter=ai-agents`, applied to the AgentCore **execution role** (`infra/lib/agent-platform-stack.ts` ~L827).
- But GLM-5 runs through **Bedrock Mantle** via a separate IAM user's **bearer token** (`AWS_BEARER_TOKEN_BEDROCK`, `openclaw.json:8`; `mantle_proxy.py:43` → `bedrock-mantle.us-east-1.api.aws`), so that spend never carries the tag → ~$0.

**Key feasibility finding:** `infra/agent-image/mantle_proxy.py` already parses **every** upstream Mantle response and reassembles SSE chunks (lines 262-306). It is the clean place to capture token `usage` from the OpenAI-compatible responses, and (because it already rewrites the request body for tool-call repair) it can inject `stream_options.include_usage=true` so streaming responses emit usage.

### Phase 0 (P0) — Capture real tokens + real model id  *(makes the usage screens correct)*

- **`infra/agent-image/mantle_proxy.py`** (capture point):
  - When `stream: true`, inject `stream_options.include_usage = true` into the forwarded request (alongside the existing tool-call-id repair) so Mantle emits a final `usage` chunk.
  - Extend the SSE reassembly loop and the non-streaming branch to read `usage.prompt_tokens` / `usage.completion_tokens` and `model` from the response.
  - Maintain module-level cumulative counters + last model; expose `GET /usage` returning them.
  - *Why cumulative + delta:* one agent turn makes many Mantle calls (tool loops). The proxy is per-container = per-microVM = per-session and turns are serial, so a before/after delta around `adapter.process()` correctly sums a single turn's usage across all sub-calls.
- **`infra/agent-image/agentcore_wrapper.py`** (plumbing, ~lines 358-375): read baseline `/usage` before `adapter.process()`, read final after, and add `input_tokens`, `output_tokens`, and the real `model` (from the proxy; fallback `zai.glm-5`) to the returned `metadata`.
- **`infra/lambdas/agent-router/index.ts:976`**: replace the stale `'kimi-k2.5'` fallback with `'unknown'` (the wrapper now always sends the real model; error paths already send `null`).
- *Result:* Stats "Total Tokens", Daily Usage, Model Breakdown (now "GLM-5 / zai.glm-5"), and per-user Adoption tokens all populate. No dashboard read-code change needed for usage.

### Phase 1 (P1) — Cost calculation + model-switch projection  *(answers both questions)*

- **Pricing in `ai_models`** (`lib/db/schema/tables/ai-models.ts`): add a row with `modelId = 'zai.glm-5'` — **must exactly match** the string the wrapper records (per `docs/guides/silent-failure-patterns.md` / `ai-model-import-guide.md`; mismatches silently yield $0) — using **researched current GLM-5 / Bedrock-Marketplace pricing** (never guessed). Ensure the candidate comparison models (Claude/Gemini/etc.) have current pricing rows.
- **New read actions** (extend `actions/admin/agent-telemetry.actions.ts` or a new `agent-cost-projection.actions.ts`):
  - `getAgentCostByModel(range)` — join per-model token aggregates against `ai_models` pricing → actual $ per model + total. Surface "pricing missing for `<model>`" explicitly instead of $0.
  - `getAgentCostProjection(range, candidateModelIds[])` — multiply the **actual** input/output token volume by each candidate model's pricing → projected $ if the agents ran on that model.
  - V1 computes cost **on read** (tokens × current pricing) — no schema/migration needed. (Note for later: a per-message cost-snapshot column would give historical accuracy across price changes.)
- **UI — `app/(protected)/admin/agents/_components/agent-cost-view.tsx`**: primary cost view from token×pricing, plus a model-switch projection table/selector. **Keep** the existing Cost Explorer panel as a secondary, clearly-labeled "AWS-billed (actual)" reconciliation view.
- **Cost Explorer reconciliation** (`actions/admin/agent-cost.actions.ts`, kept by decision): document that Mantle bearer-token spend runs under the separate IAM user `psd-agent-bedrock-<env>` (not the tagged execution role), so `costCenter=ai-agents` won't capture model spend. Relabel this panel as infra/AgentCore-cost reconciliation, and note follow-ups to investigate (activate the cost-allocation tag in Billing; determine whether Bedrock Marketplace/Mantle usage is taggable at all). The token×pricing path remains the source of truth for model cost.

## Definition of Done

- After invoking an agent on dev, `agent_messages` rows have `input_tokens` / `output_tokens > 0` and `model = 'zai.glm-5'` (verify via MCP `run_query`).
- Dashboard: Stats "Total Tokens" > 0; Model Breakdown shows "GLM-5 (zai.glm-5)" with real tokens; Adoption per-user tokens > 0.
- Cost tab shows actual $ from tokens×pricing **and** a projection of $ for ≥1 alternative model; the Cost Explorer panel is present but clearly labeled as reconciliation-only.
- `bun run lint` and `bun run typecheck` pass on the whole codebase; Python changes have unit tests for usage parsing; cost-calc action has tests (incl. the missing-pricing path).

## Verification

- **Container/usage:** deploy to dev → send an agent message via Google Chat → MCP `run_query` on `agent_messages` for non-zero tokens + `zai.glm-5`; open `/admin/agents` and confirm every tab shows non-zero data.
- **Unit-testable now:** `mantle_proxy.py` usage extraction against sample streaming + non-streaming OpenAI responses; `getAgentCostByModel` / projection math + missing-pricing path.

## Key files

- `infra/agent-image/mantle_proxy.py` — usage capture, `/usage` endpoint, `include_usage` injection
- `infra/agent-image/agentcore_wrapper.py` — read proxy usage; add tokens + real model to metadata
- `infra/lambdas/agent-router/index.ts` — fix stale model fallback (line 976)
- `actions/admin/agent-telemetry.actions.ts` (+ optional new cost-projection action file)
- `actions/admin/agent-cost.actions.ts` — relabel/clarify Cost Explorer as reconciliation-only
- `app/(protected)/admin/agents/_components/agent-cost-view.tsx`
- `ai_models` data — add `zai.glm-5` pricing row + verify candidate models priced

---

*Relates to #890 (Agent Telemetry, Admin Dashboard, and Organizational Nervous System).*

krishagel:
## Summary
Implements #1083. Fixes the agent telemetry **write side** so the admin Agents dashboard reports real GLM-5 token usage, the real model id (`zai.glm-5`, not the literal `default`), and real cost — and adds a model-switch cost projection.

### Phase 0 — capture real tokens + real model id
- **`infra/agent-image/mantle_proxy.py`**: cumulative token counters + `GET /usage`; `inject_include_usage()` (decoupled from the tool-call-repair guard); `fetch_upstream()` extracts usage + model id from streaming + non-streaming responses; only the final adopted response is counted, with per-field usage carry-forward across retries.
- **`infra/agent-image/agentcore_wrapper.py`**: before/after `/usage` delta around `adapter.process()` (trusted only when both reads succeed, else falls back to harness tokens + logs the gap); metadata carries proxy tokens + real model id (`DEFAULT_AGENT_MODEL_ID`).
- **`infra/lambdas/agent-router/index.ts`**: stale `'kimi-k2.5'` fallback → `'unknown'`.

### Phase 1 — cost calc + model-switch projection
- **Migration 088**: `ai_models` row for `zai.glm-5` — researched Bedrock pricing $1.00/1M in, $3.20/1M out; idempotent upsert.
- **`actions/admin/agent-cost-projection.actions.ts`** (new): `getAgentCostByModel`, `getAgentCostProjection`, `getPricableModels` — token×pricing, per-direction COALESCE, missing-pricing surfaced explicitly.
- **`actions/admin/agent-cost.actions.ts`**: relabeled reconciliation-only; dropped the dead `byModel` Cost Explorer query.
- **UI**: token×pricing cost view, race-guarded model-switch projection, relabeled Cost Explorer panel; shared `formatUsd` + `AGENT_MODEL_*` constants.

## Verification (all green)
- Build/Typecheck: ✅ `tsc --noEmit` clean (app + router lambda)
- Lint (zero-error): ✅ 0 errors; no new warnings in touched files
- Tests: ✅ 21 Python unittest + 12 jest (incl. missing-pricing + no-double-count paths)
- DB join verified: 40k in + 9.5k out × pricing = $0.0704 (matches UI)
- Authenticated E2E (host :3100): nexus/organization 18 passed; cost + activity pages render
- Both AI review rounds (gemini + claude[bot]) fully addressed

## Evidence
agents-cost-tab
agents-usage-tab

Closes #1083


krishagel:
## Review round 1 — all findings addressed (`eb8adc16` + `d1d149d2`)

Thanks @gemini-code-assist and Claude reviewer. Every correctness finding is fixed and re-verified.

### SQL NULL propagation (gemini, HIGH ×2) + NULL sort (gemini, HIGH)
`agent-cost-projection.actions.ts` — `getAgentCostByModel.usd`, `getAgentCostProjection.actualUsd`, and the `orderBy` now **COALESCE each direction's SUM separately** so a model priced on only one direction no longer collapses the whole cost to $0 (`X + NULL = NULL`). Verified against the local DB: with a NULL output price the input side still contributes $0.04 instead of $0.

### Baseline-failure telemetry inflation (gemini + Claude, HIGH)
`agentcore_wrapper.py` — `read_proxy_usage()` now returns an `ok` flag; the before/after delta is trusted **only when both reads succeeded**, otherwise it falls back to the harness token counts (so a mid-session proxy read failure can't attribute the container's lifetime cumulative tokens to one turn). Added a `logger.warning` when telemetry degrades so a lost read is distinguishable from genuine 0-token usage in the logs.

### Stale cost-projection race (Claude)
`agent-dashboard-client.tsx` — added a monotonic `projectionVersionRef` and a single guarded `runProjection()` used by **both** the tab loader and the candidate selector; only the latest-issued request writes state, so out-of-order resolution can't show a mismatched projection. Verified live: switching the candidate to Gemini 3.1 Pro recomputes to $0.194 (screenshot `agents-cost-projection-opus.png`).

### Test-coverage gaps (Claude)
- `getPricableModels` mapping + numeric coercion + empty-list (2 jest cases).
- `/usage` endpoint + the cumulative-counter **before/after delta contract**, asserting a single turn's delta excludes prior-turn totals — covers the "no double-counting" claim (3 unittest cases).

Gate re-run green: `tsc --noEmit` clean, `eslint` 0 errors / no new warnings, 12 jest + 52 python tests pass.

krishagel:
## Review round 2 — all findings addressed (`f56ce77b`)

Thanks for the thorough second pass. Every finding is resolved.

### Correctness
1. **Retry drops billed tokens** (`mantle_proxy.py`) — usage is now carried forward per-field: a retry's usage is adopted only when it actually reported usage (non-None), so a differently-shaped retry response no longer discards the real token count an earlier attempt generated.
2. **`inject_include_usage` coupled to the messages guard** (`mantle_proxy.py`) — decoupled: usage injection now runs on any streaming dict payload, independent of the tool-call-id-repair guard (which needs `messages`). Added a test asserting injection works with no `messages` array.
3. **Silent failure branches** (`agent-dashboard-client.tsx`) — `getPricableModels` and `getAgentCostSummary` failures now call `showError` instead of silently emptying the dropdown / reconciliation panel.

### Efficiency / dead code
4. **Dead Cost Explorer query** (`agent-cost.actions.ts`) — dropped the `USAGE_TYPE`-grouped `byModel` query (nothing renders it anymore) and removed the unused `byModel` field from `AgentCostSummary`. One fewer AWS call per load.
3 (redundant projection totals query): kept intentionally — `getAgentCostProjection` is also called standalone by the candidate selector, where the `byModel` result isn't in scope, so deriving in-memory would couple the two actions. The query is indexed on `created_at`.

### Single source of truth
5. **`zai.glm-5` hardcoded in 3+ places** — extracted `AGENT_MODEL_ID`/`AGENT_MODEL_LABEL` into `lib/agents/platform-model.ts` (kept out of the `"use server"` files, which may only export async fns), used by the projection self-exclusion + the UI "Actual" label; `DEFAULT_AGENT_MODEL_ID` in `agentcore_wrapper.py` replaces both Python fallbacks.

### Cleanup
7. **Duplicate `usd()` formatter** — extracted `formatUsd(value, maxFractionDigits)` into `lib/utils/format-currency.ts`; replaces the copies in `agent-cost-view.tsx` (precision 4) and `activity-stats-cards.tsx` (precision 2).

Gate re-run green: `tsc` clean, `eslint` 0 errors (no new warnings), 21 python + 12 jest pass.

krishagel:
🤖 **pr-fix routine round 1**

Picked up the remaining unaddressed findings from the claude-review comments posted after the round-2 fix (`5ffcec5`):

- `infra/agent-image/agentcore_wrapper.py` (`1b8eabd`) — fixed the falsy-zero bug: `input_tokens = proxy_in or result.tokens_in` discarded a real, trustworthy proxy-measured `0` and substituted the (possibly wrong) harness fallback count, because Python's `or` can't distinguish "0" from "missing." Now gated on the `usage_trustworthy` flag explicitly.
- `app/(protected)/admin/agents/_components/agent-dashboard-client.tsx` (`f042906`) — a `getPricableModels()` failure reset the candidate list to empty but left a stale `candidate` selection in place, so the projection fetch a few lines down could still run against a model no longer in the list. Now clears the selection alongside the list.
- Resolved 6 review threads that were already fixed by earlier commits in this PR (5 gemini-code-assist SQL NULL/sorting/telemetry findings from round 1, 1 codex retry-usage-carry-forward finding from round 2) but hadn't been marked resolved.

Verification: Python files compile clean (`py_compile`) and the existing `test_mantle_proxy.py` suite (20 tests) passes unchanged. **I could not run `bun run typecheck`/`bun run lint`/the jest or Playwright suites** — this sandbox's outbound access to `registry.npmjs.org` returns 403 (organization egress policy), so `bun install` cannot complete and no `node_modules` exists here. The two TS changes are small, type-safe by inspection (both branches of `ActionState` carry `.message`; `setSelectedCandidate` and `candidate` are already in scope), and mirror the exact pattern already used elsewhere in the same function — please have CI/a human confirm `tsc`/`eslint`/jest green before merge.

Not addressed (non-blocking per the reviewer bot itself, deferred as design/architectural rather than autonomously changed):
- The "turns are serial per container" invariant is still only enforced by comment, not a lock — noted as a pre-existing design gap, not a regression from this PR.
- `getAgentCostProjection`'s "actual totals" query still re-aggregates independently of `getAgentCostByModel` rather than reusing its result — a cross-action refactor with more risk than an autonomous routine should take unreviewed.




---

PR #1091, opened 2026-07-01
fix: #1090 - Atrium Phase 5 post-merge follow-ups (publish gate, route params, events, scheduler identity)

### LINKED ISSUE #1090: Atrium Phase 5 (#1055) — post-merge follow-ups from PR #1088 review
Deferred, non-blocking items surfaced by the AI review of PR #1088 (Atrium Phase 5 — Agent access). None are security bugs; all were consciously deferred to merge. Grouped by area.

## Correctness (minor)
- [ ] **Public-publish gate fires on idempotent no-op saves.** `publishService.publish()` / `visibilityService.setLevel()` compute `isPublicFacing` from the target value only, not vs. the object's current `visibilityLevel`. Re-saving already-public content (visibility unchanged) by a non-admin owner spuriously throws `ApprovalRequiredError` + emits `content.public_publish_requested`. Gate on an actual state change (current ≠ target) before throwing/emitting. Fails closed today (no exposure).
- [ ] **`extractStringParam` segment collision.** `lib/api/route-helpers.ts` `indexOf(segmentName)` misparses when a content slug is literally `"publish"` or `"content"` (e.g. `DELETE /api/v1/content/publish/publish/schoology` → wrong destination → 400, un-unpublishable). Switch to real Next.js route params, or reject slugs colliding with path literals.

## Consistency / latency
- [ ] **Fire-and-forget service-layer `contentEvents.emit`.** Still `await`ed in `version-service.ts:343`, `publish-service.ts:178,348,491`, `visibility-service.ts:529`. It swallows its own errors (`snsPublishBestEffort` never throws), so `void contentEvents.emit(...)` matches the audit-write pattern and removes an SNS round-trip from the response path (safe on long-lived ECS).
- [ ] **`use-editor-actions.ts` doesn't handle `approvalRequired`.** `handlePublish`/`handleUnpublish` branch only on `isSuccess`/`message`, unlike `VisibilityChip`. Latent — `unpublish-document.ts` is pinned to `destination: "intranet"` so the §26.4 gate can't fire today; wire the `approvalRequired` pending-approval UX before any public-destination publish/unpublish button uses this handler.
- [ ] **`lib/oauth/drizzle-adapter.ts:319` weak `ATRIUM_SYSTEM_USER_ID` validation.** Only checks `!Number.isNaN(parseInt(...))`, so `-1`/`1.5` pass. Use `systemUserIdOrNull()` (`lib/content/helpers.ts`) which requires `Number.isInteger(id) && id > 0`.

## Cleanup
- [ ] Extract a single `hasPublishPublicScope(auth)` helper — `auth.scopes.includes("content:publish_public")` is hand-typed at ~7 REST/MCP call sites.
- [ ] Dedup collection lookup on create — surface `resolveCollectionId` and service `collectionDefaultOutsideTx` both fetch the same `content_collections` row when a collection is given + visibility omitted (needs a broader `resolveCollectionId` return-type change; only worth it if that layer is refactored anyway).
- [ ] `createVersion` post-snapshot reload — avoiding it would return stale metadata under concurrent updates + drop a concurrent-delete guard; only revisit alongside a `versionService.snapshot` return-shape change.

## Known gaps carried from the PR body
- [ ] **Scheduler last mile not wired.** `app/api/assistant-architect/execute/scheduled/route.ts` accepts `agentIdentityId` but only logs a warning — it never builds an autonomous Requester. The DB column (migration 091) + Lambda forwarding are in place; the execute endpoint consuming it is not.
- [ ] **OIDC JWT signing in prod/KMS.** node-oidc-provider needs an in-process private key; KMS keys are non-exportable, so prod JWT issuance stays disabled until an exportable OIDC signing key is supplied. Autonomous authorization logic is implemented + unit-tested regardless of transport.

Refs: PR #1088, Epic #1059.

krishagel:
## Summary
Implements the deferred, non-blocking follow-ups from the AI review of PR #1088 (Atrium Phase 5 — Agent access), tracked in #1090. Correctness fixes, a latency/consistency pass, cleanup, and the scheduler identity last mile.

## Changes

**Correctness**
- **Public-publish gate no longer fires on idempotent no-op saves.** `publishService.publish()` and `visibilityService.setLevel()` computed `isPublicFacing` from the target value alone, so re-saving already-public content (visibility unchanged) by a non-admin owner spuriously threw `ApprovalRequiredError` + emitted `content.public_publish_requested`. Both now gate on an actual state change (`current visibilityLevel !== "public"`). A `public_web` destination stays unconditionally public-facing. Fails closed regardless.
- **`extractStringParam` segment-collision fixed** by threading real Next.js route params through `withApiAuth` (awaited `context.params`, passed as a 4th handler arg) and switching the 5 Atrium content dynamic routes to `params.id` / `params.destination`. Previously `DELETE /api/v1/content/publish/publish/schoology` (a content slug of `"publish"`) misparsed the destination → 400 → un-unpublishable. Backward compatible: the ~16 non-dynamic handlers ignore the new arg.

**Consistency / latency**
- **Service-layer `contentEvents.emit` is now fire-and-forget** (`void`, not `await`) in version-service, publish-service (published / unpublished / gate), visibility-service (gate), and the two content-service create gates. `emit` swallows its own errors (`snsPublishBestEffort` never throws), so awaiting only added an SNS round-trip to the response path (matches the `recordContentAudit` best-effort pattern).
- **`use-editor-actions` handles `approvalRequired`.** `handlePublish` / `handleUnpublish` now surface a §26.4 pending-approval outcome as a distinct amber "submitted for review" caption (via a new `pendingApproval` state) instead of a red failure, mirroring `VisibilityChip`. `unpublish-document` now maps `ApprovalRequiredError` like `publish-document` so the branch is reachable when a public-destination unpublish button is wired (latent today — the editor pins `intranet`).
- **`ATRIUM_SYSTEM_USER_ID` validated via `systemUserIdOrNull()`** in the OAuth adapter (was a bare `!Number.isNaN(parseInt)` that accepted `-1`, `1.5`, `"3abc"`).

**Cleanup**
- **Extracted `hasPublishPublicScope(scopes)`** — the `content:publish_public` scope check was hand-typed at 7 REST/MCP call sites (a typo would silently fail-open). Preserves the exact `.includes` semantics (a session wildcard `["*"]` still does not match).

**Known gap — scheduler last mile**
- The scheduled execute endpoint accepted `agent_identity_id` but only logged a warning. It now resolves the identity into an `agent-autonomous` Requester (`buildAutonomousRequesterForIdentity`, scopes read from the identity row since a scheduled run has no token) and **fails closed (422)** on a missing/inactive identity — no silent fallback to the user's authority. The Requester is threaded into `PromptExecutionContext` and stamped on the run log. Note: assistant-architect prompt execution currently wires only repository search tools into `promptTools`; binding the Atrium content-authoring tools (MCP-only today) to this Requester is a separate, larger feature — the tools have no prompt-execution consumer yet.

## Not addressed (documented in the issue as externally blocked)
- **OIDC JWT signing in prod/KMS** — node-oidc-provider needs an in-process private key; KMS keys are non-exportable, so prod JWT issuance stays disabled until an exportable OIDC signing key is supplied. No code change is possible here; the autonomous authorization logic is implemented + unit-tested regardless of transport.
- The remaining Cleanup bullets in the issue (`resolveCollectionId` dedup, `createVersion` post-snapshot reload) were explicitly gated in the issue on a broader refactor of those layers ("only worth it if that layer is refactored anyway") — out of scope for a follow-up pass.

## Verification (all green before opening)
- Typecheck: ✅ (`tsc --noEmit`, whole project)
- Lint: ✅ 0 errors (`eslint .`); no new warnings in touched files
- Tests: ✅ 1842 passed (`bun run test:ci` — full CI jest config, excludes e2e/perf)
- Atrium smokes: ✅ all pass (incl. the DB-backed visibility-reference smoke)
- OpenAPI: ✅ in sync (`openapi:check`) — no API contract change
- New tests: `use-editor-actions` (approvalRequired → pending, not error, for publish + unpublish), `with-api-auth` (params awaited + forwarded, using a colliding `publish` slug), `buildAutonomousRequesterForIdentity` (scopes-from-row, fail-closed on inactive id, role-less)

## Evidence
_N/A — no reachable new UI surface. The only UI change (an amber "submitted for review" caption in `DocumentEditor`) is latent: the shipped editor pins `destination: "intranet"`, which never triggers the §26.4 approval gate, so the pending-approval branch cannot be reached from the current UI. It is covered by a `renderHook` unit test (`tests/components/use-editor-actions.test.ts`) instead. All other changes are backend/service-layer._

Closes #1090

### DIFF
--- a/actions/db/atrium/unpublish-document.ts
+++ b/actions/db/atrium/unpublish-document.ts
@@ -19,6 +19,7 @@ import {
 } from "@/lib/logger";
 import { createSuccess, handleError, ErrorFactories } from "@/lib/error-utils";
 import { publishService } from "@/lib/content/publish-service";
+import { ApprovalRequiredError } from "@/lib/content/errors";
 import type { ActionState } from "@/types";
 import { hasCapabilityAccess } from "@/utils/roles";
 import { getServerSession } from "@/lib/auth/server-session";
@@ -74,6 +75,21 @@ export async function unpublishDocumentAction(
     return createSuccess(result, "Document unpublished");
   } catch (error) {
     timer({ status: "error" });
+    // §26.4 gate: taking a public destination offline without authority is a
+    // pending-approval outcome (approval-queue event emitted in the service), not a
+    // failure — surface it distinctly so the editor can show an amber "submitted
+    // for review" caption, mirroring publishDocumentAction. Defensive: the shipped
+    // editor only unpublishes from `intranet` (which never trips the gate), but the
+    // service accepts public destinations.
+    if (error instanceof ApprovalRequiredError) {
+      log.info("Unpublish requires approval", { requestId });
+      return {
+        isSuccess: false,
+        approvalRequired: true,
+        message:
+          "Unpublishing from this destination requires administrator approval — your request has been submitted for review.",
+      };
+    }
     return handleError(error, "Failed to unpublish document", {
       context: "unpublishDocumentAction",
       requestId,
--- a/app/api/assistant-architect/execute/scheduled/route.ts
+++ b/app/api/assistant-architect/execute/scheduled/route.ts
@@ -9,6 +9,8 @@ import { promptResults, scheduledExecutions } from '@/lib/db/schema';
 import { unifiedStreamingService } from '@/lib/streaming/unified-streaming-service';
 import { retrieveKnowledgeForPrompt, formatKnowledgeContext } from '@/lib/assistant-architect/knowledge-retrieval';
 import { createRepositoryTools } from '@/lib/tools/repository-tools';
+import { buildAutonomousRequesterForIdentity } from '@/lib/content';
+import type { Requester } from '@/lib/content/types';
 import type { StreamRequest } from '@/lib/streaming/types';
 import type { UIMessage } from 'ai';
 import jwt from 'jsonwebtoken';
@@ -69,6 +71,14 @@ interface PromptExecutionContext {
   userCognitoSub: string;
   assistantOwnerSub?: string;
   userId: number;
+  /**
+   * The autonomous agent Requester the run executes as (§25 service identity),
+   * when the schedule specified an `agent_identity_id`. Null = the run authors any
+   * Atrium content as the owning user. Threaded here so the content-authoring tools
+   * (a later wiring) resolve their write permission from the identity's scopes, not
+   * the user's full human authority.
+   */
+  agentRequester: Requester | null;
 }
 
 /**
@@ -110,6 +120,54 @@ function validateInternalRequest(req: NextRequest): { scheduleId: string; execut
   } catch (error) {
     log.warn('JWT verification failed', { error: error instanceof Error ? error.message : String(error) });
     return null;
+  }
+}
+
+/**
+ * Resolve the service-identity Requester for a scheduled run (§25). Returns the
+ * `agent-autonomous` Requester when `agentIdentityId` is set and active, `null`
+ * when the schedule runs as the owning user (no identity), or an error `Response`
+ * (HTTP 422) when a requested identity is missing/inactive — the run FAILS CLOSED
+ * rather than silently executing with the user's authority.
+ *
+ * Extracted from `POST` to keep that handler within the complexity/line budget.
+ */
+async function resolveScheduledAgentRequester(
+  agentIdentityId: string | null | undefined,
+  ctx: { scheduleId: number; userId: number; requestId: string; timer: ReturnType<typeof startTimer>; log: ReturnType<typeof createLogger> }
+): Promise<{ requester: Requester | null } | { errorResponse: Response }> {
+  if (!agentIdentityId) return { requester: null };
+  try {
+    const requester = await buildAutonomousRequesterForIdentity(agentIdentityId);
+    ctx.log.info('Scheduled run resolved to an agent service-identity', {
+      scheduleId: ctx.scheduleId,
+      userId: ctx.userId,
+      agentIdentityId,
+      agentLabel:
+        requester.kind === 'agent-autonomous' ? requester.agentLabel : undefined,
+    });
+    return { requester };
+  } catch (identityError) {
+    ctx.log.error('Scheduled run requested an inactive/unknown agent identity; aborting', {
+      scheduleId: ctx.scheduleId,
+      userId: ctx.userId,
+      agentIdentityId,
+      error:
+        identityError instanceof Error ? identityError.message : String(identityError),
+    });
+    ctx.timer({ status: 'error', reason: 'invalid_agent_identity' });
+    return {
+      errorResponse: new Response(
+        JSON.stringify({
+          error: 'Invalid agent identity',
+          message:
+            'The schedule references an agent identity that is not active. ' +
+            'Re-activate it or clear the identity to run as the owning user.',
+          requestId: ctx.requestId,
+        }),
+        { status: 422, headers: { 'Content-Type': 'application/json' } }
+      ),
+    };
   }
 }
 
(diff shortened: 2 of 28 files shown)


krishagel:
Thanks @gemini-code-assist — addressed all three in d775c79e.

Note: `agent_identities.scopes` is `text[].notNull` (schema), so these were defensive rather than live bugs, but the guards are cheap and make a malformed/legacy row or test double unable to crash an authority check:

- **requester-from-auth.ts:186** — `scopes: identity.scopes ?? []`.
- **helpers.ts:239** — `hasPublishPublicScope(scopes: string[] | null | undefined)` now uses `scopes?.includes(...) ?? false` (a malformed context DENIES, fail-safe).
- **requester-from-auth.ts:127** — kept the loader return type as `string[]` (accurate to the `notNull` column); the `?? []` at the build site is the runtime safety net, so no downstream null-handling is forced.

Added unit tests for `hasPublishPublicScope` covering explicit-scope match, wildcard-does-not-match (no fail-open), and null/undefined/empty safety.

krishagel:
Round 2 addressed in b2904d9f.

**The converged P2 (TOCTOU on the §26.4 gate)** — real bug, fixed. Thanks @chatgpt-codex-connector / @Copilot / @gemini-code-assist for catching it. The idempotency check I added read the current visibility level *before* the `FOR UPDATE` lock, so a concurrent narrow (public→internal) could let an unauthorized widen-back-to-public skip approval. The 'is this a NEW public exposure?' decision now runs **inside the transaction against the locked row**:
- `publishService.publish`: split the gate — the `public_web`-destination branch stays pre-tx (level-independent, and must precede the adapter-not-implemented check so an unauthorized caller still gets the approval signal); the visibility-widen branch moved into the tx against `locked.visibilityLevel`.
- `visibilityService.setLevel`: reads the current level under the same lock that guards the write and gates there. This also fixes the Copilot null-on-concurrent-delete case (the lock lookup 404s first).

**Minor:**
- `helpers.ts` doc — corrected: a scope-string typo fails **closed**, not open (Copilot was right).
- `with-api-auth` `RouteParams` values are now `string | undefined` so handlers null-check a non-segment key (routes already guard `if (!id)`).

Added race-safe idempotency tests (no-op re-save/re-publish of already-public content passes without approval; a genuine widen from a non-public *locked* level is gated). 1847 CI tests green.

krishagel:
Round 3 — the three comments re-posted on b2904d9 are all already-addressed (the bots re-scan the full diff each push and don't auto-close resolved threads):

- **requester-from-auth.ts:189** (`scopes: identity.scopes ?? []`) — already applied in d775c79e; it's on the current line.
- **requester-from-auth.ts:127** (loader return type) — intentionally kept `string[]` (accurate to the `text[].notNull` column); the `?? []` at the build site is the runtime net, so no downstream null-handling is forced.
- **visibility-service.ts:537 (Copilot, 'gate before FOR UPDATE')** — this is exactly what b2904d9f fixed. The gate now runs **inside** `executeTransaction`, reads `rows[0].visibilityLevel` under the `FOR UPDATE` lock (lines ~549–563), and throws before `setLevelInTx`. Line 537 is now just the `mayPublishPublic` authority computation + a comment explaining why the decision is deferred into the tx. No pre-lock gate remains.

No outstanding actionable items. CI green (CodeQL, claude-review, Test/Lint/TypeCheck, CDK), 1847 tests passing, mergeable.

krishagel:
🤖 **pr-fix routine — round 1**

Addressed the claude[bot] review findings from 18:35 (all other outstanding review threads — gemini/codex/copilot rounds 1–2 — were already fixed by @krishagel in `d775c79e`/`b2904d9f`, confirmed by re-reading the current code):

1. **`tests/unit/atrium-publish-service.test.ts`** — added `expect(setLevelInTxCalls).toBe(0)` to the visibility-widen rejection test, so a future reorder that ran the write before the gate would fail the test, not just throw the right error type.
2. **`tests/unit/atrium-public-visibility-gate.test.ts`** — added an equivalent `txUpdateCalls` counter (tracks `tx.update(...)` specifically, distinct from the `.limit()` lock read) and asserted it stays 0 on all three rejected-`setLevel` tests. Also corrected the file's header comment, which still described the pre-#1090 ordering ("gate runs before any DB write") after the race-free rewrite moved the gate inside the transaction.
3. **Extracted `raisePublishApprovalRequired`** into `lib/content/helpers.ts` so `publishService` and `visibilityService.setLevel` share one emit-then-throw implementation of the §26.4 gate instead of `visibility-service.ts` hand-writing its own copy inline (payload shapes intentionally differ — publish includes `slug`/`destination`, setLevel doesn't — preserved exactly).
4. **`components/atrium/use-editor-actions.ts`** — extracted `mapActionResult` to remove the duplicated `isSuccess`/`approvalRequired`/else branching between `handlePublish`/`handleUnpublish`.
5. **`ActionOutcome`** is now a 3-state discriminated union (`success`/`pending`/`error`) instead of `{ok: boolean; pending?: boolean}`, removing the double-negative in `runAction` and making the invalid `{ok: true, pending: true}` combination unrepresentable.

**Not addressed (declining, with reasoning):**
- **Finding #6** (`with-api-auth.ts` 4th positional param / ~15 other routes still on `extractStringParam`) — the review itself frames this as "worth a follow-up issue," not a change for this PR. Leaving as-is; happy to pick it up if a follow-up issue is opened.
- **Finding #7** (`Promise.all` for `resolveScheduledAgentRequester` + `getUserById`) — parallelizing would add an unconditional `getUserById` DB call even on the fail-closed 422 path (today it short-circuits before that query). The review itself flags this as "low severity... scheduled executions aren't latency-sensitive," so I left the sequential short-circuit behavior intact rather than trade a rare-path DB call for a latency win nobody needs here.

Pushed as `4dea502` (rebased onto `787321d`, a learnings-capture commit that landed on the branch after I started). Note: I could not run `bun install`/`tsc`/`jest` locally in this sandbox (registry.npmjs.org access is blocked), so these changes are verified by careful manual reasoning + a standalone TS syntax parse, not a live test run — please keep an eye on CI.




---

Issue #1059, opened 2026-06-24
Epic: Atrium — Agent-native content workspace & intranet


krishagel:
## Context

Atrium is a **content layer on top of AI Studio** where agents and people co-create two kinds of content — **documents** (markdown edited collaboratively, per-character provenance) and **interactive artifacts** (sandboxed HTML/JS, per-version provenance) — version them with visible authorship, and publish them at controlled access levels to a staff intranet, to families, and to external systems.

**The one non-negotiable architectural commitment:** the `lib/content/` **service layer is the source of truth** for how content is created, versioned, and published. Every surface is a thin client of it — the in-app editors beside Nexus chat, external agents over MCP, scripts over REST v1, and scheduled skill runs. **There is no UI-only creation path.** This is what makes "how do agents outside Nexus create content?" a non-question: they use the same API the editors use.

Full design spec: `docs/features/atrium-design-spec.md` (1620 lines, implementation-ready). This epic decomposes it into 9 phases (0–8). **Phase 8 (OKF interoperability) was added after the spec was written** and is captured there as **§36**.

## Goals

- **Parity** — anything a person can do through the UI, an agent can do through tools.
- **Granularity** — tools are atomic primitives (`create`, `update`, `publish`); behaviors are outcomes an agent composes, not bundled workflow code.
- **Composability** — new capabilities arrive as new prompts/skills over the same primitives, not new endpoints.
- **Content as context** — published content is legible to agents and retrievable as grounding, scoped by permission.
- **Service layer is truth** — editors, MCP, REST, and schedules all call `lib/content/`.

## Vocabulary (committed)

- **Content object** — the addressable unit. `kind`: `document` or `artifact`.
- **Document** — markdown edited in Proof, rendered rich via templates; per-character provenance (green=human, purple=agent).
- **Artifact** — interactive HTML/JS edited in a canvas, rendered in a cross-origin sandbox; per-version provenance.
- **Collection** — an intranet section; also the default-visibility grouping and a retrieval scope.
- **Visibility** — who may consume: `private` / `group` (role·building·department·grade·user) / `internal` (any authenticated) / `public` (anonymous).
- **Destination** — where surfaced: `intranet` / `public_web` / `schoology` / `google`.
- **Delegated agent** — acts on behalf of a user, inheriting exactly their permissions.
- **Autonomous agent** — a service/skill identity with its own role + scopes; **cannot publish publicly**.

## Architecture (4 layers — only the content layer is built from scratch)

[...]

PR #1108, opened 2026-07-04
feat(atrium): Phase 6 permission-aware retrieval — content as scoped context (#1056)

### LINKED ISSUE #1056: Atrium Phase 6 — Retrieval (content as scoped context)
**Phase:** 6 of 7 · **Depends on:** Phases 0, 3, 5 (events) · Part of **Epic #1059** (Atrium content workspace).

## Context

Closes the "content as context" loop: published content becomes **permission-aware** retrieval grounding for assistants. Every search hit is filtered by `canView`, so the same content store safely serves a staff assistant and a student assistant differently — the structural guarantee against leaking restricted content via RAG. See Epic #1059 for the full context.

## Summary

Make published content retrievable as **permission-aware** scoped context for assistants — index on publish via the existing repository pipeline, and filter every search hit by `canView` (the safety boundary).

## Deliverables

- **Index on publish / new version** (`lib/content/retrieval-service.ts`, §16.1): on `content.published`, register/refresh a `repository_item`, chunk + embed, link via `content_index_links`. Store filterable retrieval metadata (objectId, kind, collectionId, visibilityLevel, tags, grants mirror); stamp `content_objects.indexed_at`.
- **Permission-aware `search`** (§16.2): run existing semantic search, then drop any hit whose object fails `canView(req, obj)` — never return what the requester can't see. Enforced server-side, never by UI filtering.
- **Whole-object injection** (§16.3): `getContextDocument(objectId)` returns full markdown/text verbatim (the `context.md` pattern).
- **Assistant retrieval scoping** (§16.4): an Assistant Architect assistant stores a scope (`collectionId` and/or `tags`, optional max visibility tier); scope narrows candidates and `canView` enforces per-requester access — same store safely serves staff and student assistants.

## Reuse (verified existing code)

- Retrieval stack: `repository_items`, `repository_item_chunks` (pgvector `vector()` custom type), `lib/repositories/search-service.ts` (vector/keyword/hybrid), `generateEmbedding`.
- Phase 3 `canView`; Phase 5 `content.published` event as the index trigger.

## Gaps / risk callouts

- **Open decision (spec §33 #4):** reuse `knowledge_repositories`/`repository_item_chunks` with a system repository per collection (recommended) vs a dedicated content index; reconcile `repository_access` vs `content_visibility_grants`. This spec filters by `canView` at query time regardless.
- The `canView`-per-hit filter is the structural guarantee a student-facing assistant cannot retrieve staff-only content — must be load-tested for N hits.

## Acceptance criteria (spec §31.2, Phase 6)

- [ ] Published content is indexed.
- [ ] Permission-aware `search` excludes restricted content for the wrong requester (integration test: staff-scoped assistant retrieves a doc; student-scoped assistant does not).
- [ ] `bun run lint` + `bun run typecheck` pass.

📄 **Full spec (committed):** `docs/features/atrium-design-spec.md` — read the sections referenced below.

## Spec references

§5 flow D, §16 (retrieval), §28.2 (permission-aware retrieval), §33 #4.

krishagel:
Closes #1056 · Part of Epic #1059 (Atrium content workspace) · Phase 6 of 7

## Why this was `lfg-blocked`

Issue #1056 was labeled `lfg-blocked` **not** because of a code problem — the cloud lfg routine's network policy blocks `bun install` (`403 host_not_allowed` on every package host), so the Definition-of-Done gate (build/lint/typecheck/tests) was entirely unexecutable there. The routine wrote a full implementation, then discarded it unverified and pasted the diff into the issue.

This PR applies that work **locally where the gate can run**, corrects it against the actual codebase (every referenced symbol was verified before applying — several signatures the routine guessed at were confirmed real), adds the required integration test, and runs the full gate green.

## What this does (spec §16 / §28.2)

Published content becomes **permission-aware** retrieval grounding for assistants. Every search hit is filtered by `canView`, so the same content store safely serves a staff assistant and a student assistant differently — the structural guarantee against leaking restricted content via RAG.

New `lib/content/retrieval-service.ts`, reusing the existing repository/vector pipeline (`knowledge_repositories` / `repository_items` / `repository_item_chunks` / `lib/repositories/search-service.ts`) rather than a parallel index (spec §33 #4):

- **`indexObject` (§16.1)** — chunk + embed the current version's text, upsert the backing `repository_item`/chunks in one shared system-owned repository, link via `content_index_links` (Phase 0), stamp `content_objects.indexed_at`. Embeddings generated **before** the transaction so a slow provider call never holds a DB tx open. Idempotent re-index on new versions.
- **`search` (§16.2, the safety boundary)** — vector search scoped to the Atrium repo → resolve each hit to its object → narrow by scope → **drop any hit whose object fails `canView(req, obj)`**, re-checked against a freshly loaded object (never trusting mirrored metadata for a permission decision). Server-side, never UI filtering.
- **`getContextDocument` (§16.3)** — whole-object injection: full markdown/source verbatim, or `null` if not published / not viewable.
- **`searchForAssistant` (§16.4)** — loads the assistant's stored `retrievalScope` and runs a scoped search; `canView` still enforces per-requester access on top of the scope.

### Index-on-publish wiring
`publishService.publish` now calls `retrievalService.indexObject` synchronously (best-effort; failures logged + swallowed) right after the commit, so a successful publish guarantees "published content is indexed". The `content.published` SNS event has **no retrieval subscriber** wired (events-bus infra gap), so the inline call is what actually makes the AC true. Post-commit side effects (index + event emit) were extracted into `runPublishSideEffects` to keep `publish` within its method-length budget.

### Assistant scope storage
- Migration `094-atrium-retrieval-scope.sql`: `assistant_architects.retrieval_scope` jsonb (`ADD COLUMN IF NOT EXISTS`, registered in `migrations.json`). **Applied + verified against local postgres** (`retrieval_scope | jsonb`).
- `AssistantRetrievalScope` type + `retrievalScope` column on the Drizzle schema, threaded through the `getAssistantArchitects*` queries + the two action-layer builders that reconstruct the full `SelectAssistantArchitect` shape (required for the schema addition to typecheck end-to-end).

## Acceptance criteria (spec §31.2, Phase 6)

- [x] **Published content is indexed** — `indexObject` wired into publish; test-covered.
- [x] **Permission-aware `search` excludes restricted content for the wrong requester** — integration test: a **staff** requester retrieves a group-scoped doc; a **student** requester does not.
- [x] **`bun run lint` + `bun run typecheck` pass** — only a pre-existing `max-lines-per-function` advisory on `publish` remains (158 code-lines on clean `dev` before this change; this PR is length-neutral).

## Verification

- **Tests**: `tests/unit/atrium-retrieval-permission-aware.test.ts` — 13 cases exercising the **real** `visibilityService.canView` + `helpers.principalOf` (grants driven through the DB mock), covering the staff-vs-student safety boundary, scope narrowing, whole-object gating, and index writes. All 21 Atrium suites / 81 real unit suites (1202 tests) pass.
- **typecheck**: clean (only pre-existing `.next/types` `RouteContext` noise, present on `dev`).
- **lint**: no errors; one pre-existing `max-lines` advisory (unchanged).
- **migration**: applied to local postgres, column verified.

## Known follow-up (not in scope)

`getAtriumRepositoryId` caches the system repository per process with no UNIQUE constraint on its name, so a first-ever-publish race across instances could create two system repos (narrow window, low impact). Clean fix: UNIQUE-name constraint + `ON CONFLICT DO NOTHING`.

_Generated with assistance from Claude Code._


krishagel:
## Review findings — resolution (commit `b1fe25b7`)

Thanks to the review bots. Triaged each finding on merit:

### Fixed
| Finding | Source | Resolution |
|---|---|---|
| **`repository_items.type` CHECK rejects `"artifact"`** — artifacts publish but silently never index (best-effort swallows the insert error) | Codex P2 | **Confirmed against the real CHECK on local postgres.** Map `document → "document"`, `artifact → "text"` (an artifact is indexed as its extracted source text). Regression test added. |
| **`searchForAssistant` runs UNSCOPED when the assistant id is missing** | Gemini (security-high) | Fail closed — return `[]` when no row. (`canView` already gated every hit, so this was not an authz leak, but it should not widen beyond the intended scope.) Test added. |
| **`retrieval_scope` had no write path** (column was un-settable via app code) | Gemini (high) | Threaded through `AssistantArchitectData`/`AssistantArchitectUpdateData`, `createAssistantArchitect`, and the action-layer `AssistantArchitectBaseUpdates` + builder (explicit `null` clears; absent = untouched). |
| **`indexObject` assumes 1 embedding per chunk** | Copilot | Throw on a count mismatch rather than writing `undefined` embeddings that `vectorSearch`'s `embedding IS NOT NULL` filter would silently drop. |
| **Docstring implies indexing is an invariant** | Copilot | Reworded `runPublishSideEffects` — indexing is best-effort (caught + logged), not a guarantee. |

### Deliberately not changed (with rationale)
- **N+1 in `search` (per-hit `canView`)** — Gemini: the per-hit `canView` **is** the documented safety boundary (§16.2), bounded by the search limit (default 10). The suggested `getUserCapabilities()` set does not apply here — capabilities gate UI features, not content visibility; there is no pre-fetchable set of viewable object ids. A batched grants pre-load is a viable future optimization but would restructure the safety boundary; correctness-first is the Phase 6 mandate.
- **`getAtriumRepositoryId` concurrent-create race** — Copilot: real but narrow (first-ever-publish window only). Tracked as a follow-up in the PR body; a global `UNIQUE` on the shared `knowledge_repositories.name` is a broader, riskier change than the window warrants.
- **`obj.tags` null guard** — Gemini: **false positive.** The DTO mapper coerces `tags` to `[]` (`mappers.ts`: `row.tags ?? []`), so `obj.tags` is never null at runtime.
- **Test embeddings 3-dim vs `pgvector(1536)`** — Copilot: the test fully mocks the write transaction, so `toDriver` dimension validation never runs in-test regardless of vector length. Real dimension safety comes from reusing the same embedding + column pipeline as existing repositories.

Full real unit suite (81 suites / 1204 tests) green; typecheck clean; migration applied + verified on local postgres.

krishagel:
## Round-2 review — resolution (commit `ffc011ab`)

Strong catch on the blocking finding — that's a genuine cross-component gap unit tests couldn't have surfaced. Verified it against the code and fixed it.

### 🔴 Blocking: `canView` bypass via generic repository actions — **fixed**
Confirmed: `searchRepository` (`actions/repositories/search.actions.ts`) checks only `getServerSession()`, takes a client-supplied `repositoryId`, and returns raw chunk `content` — so any authenticated user could search the Atrium system repo's (guessable serial) id and retrieve content for objects they can't `canView`.

Fix (your option B — exclude the system-managed repo from the generic API):
- The Atrium repo is now created with `metadata: { systemManaged: true }`.
- New pure guard `isSystemManagedRepository(repo)` (re-exported via the drizzle barrel).
- Every generic read path that takes a `repositoryId` now refuses a system-managed repo, **masked as not-found** so the id is not enumerable: `searchRepository` (the any-user hole), `getRepository`, `listRepositoryItems`, `searchRepositoryItems`.
- The exclusion is at the **action layer**, not in `search-service` — `retrievalService.search` calls `vectorSearch` directly by id and re-checks `canView`, so legitimate Atrium retrieval is unaffected while the generic API can no longer reach the index.
- New test `atrium-retrieval-system-repo-guard.test.ts` proves the real guard **runs no search** (no content leak) for a system-managed/non-existent repo, and that a normal repo still searches.

### 🟢 `withinScope` coverage — **fixed**
Added dedicated tests for the `tags` and `maxVisibilityLevel` branches (only `collectionId` was covered before).

### Nit: `maxVisibilityLevel` semantics — **fixed**
Clarified the JSDoc: it caps how *broadly exposed* a candidate may be (`private < group < internal < public`) — an upper bound on exposure, not a floor on restriction.

### 🟡 Publish blocks on embedding generation — **kept synchronous, by design**
Deliberately awaited: the AC (§31.2) is "published content **is** indexed", and there is currently no background-job/subscriber infra (the `content.published` SNS event has no consumer). Fire-and-forget would weaken the guarantee to eventually-consistent with no retry path. The latency is bounded (chunk count × embedding round-trip) and the call is best-effort (failures caught + logged). A background-worker offload is the right future optimization once the events bus has a subscriber — noted alongside the existing `getAtriumRepositoryId` race follow-up.

### Not changed: pre-existing generic-repository IDOR
`searchRepository` enforces no repo-level authz for *any* repository (not just Atrium) — a pre-existing issue independent of this PR. This change closes the Atrium-specific bypass; a full rewrite of the generic repository authorization model is out of scope here and worth its own issue.

Full real unit suite (82 suites / 1212 tests) green; typecheck clean.

krishagel:
## Round-3 review — resolution (commit `9cfc3176`)

You were right that round-2's guard wasn't applied everywhere the shared tables are reachable. Closed the remaining paths — the system-managed Atrium repo is now unreachable through **every** generic / admin / assistant-execution repository path.

### 🔴 Blocking: `adminUpdateRepository` could reopen the bypass — **fixed**
An admin flipping `isPublic: true` (or stripping the `systemManaged` flag) on the Atrium repo would have re-opened the exact hole. The Atrium repo is now **immutable** through both `adminUpdateRepository` and the generic `updateRepository` — masked as not-found. Tested: `adminUpdateRepository` on the system repo performs **no write**.

### Admin direct-read exposure — **fixed**
`adminGetRepositoryItems` and `listAllRepositories` no longer expose the raw indexed Atrium chunks/metadata. Tested: `adminGetRepositoryItems` reads **no items** for the system repo.

### Assistant-execution RAG paths (defense-in-depth) — **fixed**
- New reusable `EXCLUDE_SYSTEM_MANAGED` SQL predicate added to the three access-check functions (`getUserAccessibleRepositories`, `getAccessibleRepositoryIds`, `getAccessibleRepositoriesByCognitoSub` — the last is what `repository-tools` resolves access through).
- `retrieveKnowledgeForPrompt`'s raw-SQL access query now excludes `(metadata->>'systemManaged') IS DISTINCT FROM 'true'`.

So even if immutability were somehow bypassed, these paths can never resolve the system repo as accessible. `retrievalService.search` calls `vectorSearch` directly by id (re-checking `canView`) and is unaffected — the exclusions live at the action / access-check layer, never in the shared search-service.

### Coverage
New `atrium-system-repo-admin-guard.test.ts` + the round-2 `atrium-retrieval-system-repo-guard.test.ts` now assert **no write / no read / no search** occurs for a system-managed repo across the generic and admin paths. Full real unit suite (83 suites / 1216 tests) green; typecheck clean.

### Still deferred (with rationale, unchanged)
- `getAtriumRepositoryId` first-publish race — narrow window, low impact; UNIQUE-name + ON CONFLICT is the clean follow-up.
- Synchronous embedding on publish — kept awaited for the "is indexed" AC; background-worker offload is the future optimization once the events bus has a subscriber.
- The pre-existing generic-repository IDOR over ordinary uploaded documents (`searchRepository` has no repo-level authz for *any* repo) — separate, pre-existing, worth its own issue.

krishagel:
## Round-4 review — resolution (commit `c90e4435`)

Thanks — glad the security-conscious iteration held up under a fresh full-diff pass. Addressed the nit; the rest are tracked follow-ups.

- **Nit: `EXCLUDE_SYSTEM_MANAGED` defined between imports** — **fixed**, moved into the Constants section below all imports (no behavior change).
- **Pre-existing `searchRepository`/`listRepositoryItems`/`searchRepositoryItems` per-repo authz gap** — agreed this is worth its own issue. It predates this PR and isn't regressed here (the system-managed guard this PR adds is orthogonal). Will surface it as a follow-up rather than expand this PR's scope into a rewrite of the generic repository authorization model.
- **`getAtriumRepositoryId` race** — deferred as noted; UNIQUE-name + `ON CONFLICT` is the clean fix.
- **Duplicated "fetch → check system-managed → mask" block (×4-5)** — acknowledged. Left inline because the sites differ in error handling (`searchRepository` returns an `ActionState` message; the others throw) and a shared helper in the drizzle layer would couple it to the action-layer error types; each site has direct test coverage. Open to consolidating if preferred.
- **`search()` sequential `canView` loop** — kept sequential: it's bounded by the search limit (default 10) and the object-load cache dedups multi-chunk hits from the same object; parallelizing the safety-critical loop adds complexity for marginal gain at this scale. Noted as a future optimization.

CI green, mergeable. Migration is additive (`ADD COLUMN IF NOT EXISTS`), applied + verified on local postgres.

krishagel:
## Round-5 review — resolution (commit `896ce9d2`)

Excellent catch on `getItemChunks` — that was a real content-leak path I'd missed because it's keyed by **item id**, not repository id, so the round-2/3 repositoryId guards didn't cover it.

### 🔴 `getItemChunks` — direct chunk-content read — **fixed**
It returned raw `repository_item_chunks.content` after only a generic capability check. Now guarded via item→repo resolution. Test proves it performs **no chunk read** for a system-managed item.

### Also closed the rest of the class
- **Content/download by item id**: `getDocumentDownloadUrl` guarded.
- **Delete paths (🟠)**: `removeRepositoryItem`, `adminRemoveRepositoryItem` (item id), `deleteRepository`, `adminDeleteRepository` (repo id) — all now refuse system-managed targets (out-of-band delete would desync the index).

### Centralized the guard (addresses the round-4 duplication note)
New `lib/repositories/system-repo-guard.ts`:
- `assertNotSystemManagedRepository(repositoryId)`
- `assertItemNotInSystemManagedRepository(itemId)` (resolves item → repo)

Every prior inline guard now delegates to it; the admin-local `assertGenericRepository` was removed. (`searchRepository` keeps its inline form since it returns an `ActionState` message rather than throwing.)

**Net**: the Atrium system repo is now unreachable — read, search, download, mutate, or delete — through every generic/admin repository action by either **repository id or item id**. `retrievalService` (per-hit `canView`) is the only reader.

### Deliberately not guarded (rationale)
The add-item paths and `updateItemProcessingStatus` write/mutate but can't **leak** content, are ownership-gated (Atrium repo is system-owned), and any orphan rows wouldn't surface in `retrievalService.search` (resolved via `content_index_links`).

Full real unit suite (84 suites / 1220 tests) green; typecheck clean.

krishagel:
## Round-6 review — acknowledged; converged

Thanks for the thorough final pass and for confirming the system-repo isolation covers every read path into the shared tables. Agreed the four remaining points are follow-ups for later Atrium phases, not changes to this PR — capturing them here so they're tracked:

1. **`search()` over-fetch** — `canView` filters after the vector `LIMIT`, so a narrow-visibility requester can under-return. Fails closed (never leaks), but degrades result quality. The right fix (over-fetch factor or pagination) depends on how the consumer paginates, so it belongs with the phase that wires `searchForAssistant` into assistant execution. Deferred.
2. **`searchForAssistant` / `getContextDocument` have no callers yet** — confirmed intentional: this is Phase 6 of 7 (index-on-publish + the permission-aware retrieval API). Consumption by the assistant-execution path is a later phase; this is deliberate plumbing, not a missed wiring step.
3. **No zod validation on `retrievalScope`** — correct, it's raw JSONB plumbing with no client-facing caller today. Validation (shape + `maxVisibilityLevel` enum) belongs with the phase that exposes it to the assistant-editor UI. Deferred.
4. **`indexObject` `content_index_links` upsert race** — same category as the disclosed `getAtriumRepositoryId` race; low severity (best-effort, rarely-concurrent republish). Tracked with that follow-up.

No code changes this round. CI green, mergeable; migration additive and verified on local postgres.

krishagel:
## Folded in the pre-existing generic-repository IDOR fix (commit `b2d86985`)

Per request, closing the broader IDOR the earlier rounds flagged (not just the Atrium-specific slice).

**The hole**: the generic repository READ actions authorized only `getServerSession()` (or the blanket `knowledge-repositories` capability) — never whether the caller can access the *specific* repository. With sequential integer ids, any authenticated user could read a private repo they don't own / aren't granted:
- `searchRepository` (no capability at all) → raw chunk `content`
- `listRepositoryItems` / `searchRepositoryItems` / `getRepository` → item + repo metadata
- `getItemChunks` / `getDocumentDownloadUrl` (by item id) → chunk content / download URL

**Fix**: each now requires access via the canonical `getAccessibleRepositoriesByCognitoSub` model (public / owner / `repository_access` grant — the same check `repository-tools.ts` uses), via two shared guards in the renamed `lib/repositories/repository-access-guard.ts`. This **subsumes** the system-managed guard on read paths (the access query already excludes system-managed repos), so one check closes both the generic IDOR and Atrium isolation. Write/delete paths keep `assertNotSystemManagedRepository` (they gate on ownership).

**No regression**: the repository UI lists via `getRepositoriesByOwnerId` (owner-only) and navigates by id, so callers always pass accessible repos; the detail page already `notFound()`s on `!isSuccess`, so an inaccessible id now 404s (existence-masked) instead of leaking. Admins keep the separate admin actions for cross-repo access.

Tests reworked for the access model; full real unit suite (84 suites / 1219 tests) green; typecheck + lint clean.

Issue #727, opened 2026-02-04
investigate: Review first-day Bedrock guardrail violations in production


krishagel:
## Summary

Bedrock guardrails went live in production on 2026-02-03. Within the first ~12 hours, 5 violations were logged. Analysis of CloudWatch logs reveals the violations break down into two categories with different root causes and action items.

### Violation Timeline (all times UTC, 2026-02-03/04)

| # | Time | Category | Route | User ID | Module | Context |
|---|------|----------|-------|---------|--------|---------|
| 1 | 15:42:47 | **Self-Harm** | `api.nexus.chat` | 232 (staff) | Nexus Chat | User had uploaded `KS_Hands_to_Self_Graph.pdf` attachment. 3 messages in conversation using GPT-5. Content length: 341 chars. |
| 2 | 16:27:22 | **Prompt attack** | `api.nexus.chat` | 80 (staff) | Nexus Chat | User uploaded `danielson-2022-rubrics-wa-criteria-critical-attributes.pdf`. First message in conversation, using GPT-4o. Prompt started with "as an expert, veteran principal in the s..." |
| 3 | 19:59:48 | **Prompt attack** | `api.assistant-architect.execute` | 86 (staff) | Assistant Architect | Tool #100 "Danielson Observation Helper (STILL IN DRAFT)", prompt #101 "Danielson Observation Analysis", using Claude Haiku. Execution #396. |
| 4 | 20:00:00 | **Prompt attack** | `api.assistant-architect.execute` | 86 (staff) | Assistant Architect | Same user, same tool/prompt, retry 12 seconds later. Execution #397. |
| 5 | 00:16:28 | **Prompt attack** | `api.nexus.chat` | 38 (staff) | Nexus Chat | Conversation title: `**SYSTEM INSTRUCTION** You are an expert...`. Using Claude Haiku. This is a genuine prompt injection attempt. |

### Analysis

**Violation #1 (Self-Harm) - FALSE POSITIVE:**
- A staff member uploaded a PDF called `KS_Hands_to_Self_Graph.pdf` — this is a behavior tracking document (keeping hands to self, a common K-12 PBIS behavior goal)
- The topic-based Self-Harm filter likely triggered on the phrase "hands to self" or similar behavioral language in combination with the document content
- This is a **false positive** — the content is legitimate K-12 behavioral data
[...]

PR #1131, opened 2026-07-06
fix: Nexus chat production outage (GUARDRAIL_HASH_SECRET) + workspace panel layout collapse

krishagel:
## Summary

Two fixes for the post-epic Nexus breakage, both reproduced and verified with Playwright + a local **webpack production build** (deployed-dev conditions):

1. **Every Nexus chat request 500'd on deployed dev** — missing `GUARDRAIL_HASH_SECRET`.
2. **Workspace panel document layout collapse** — the editor crushed to a ~80px sliver beside a viewport-keyed comment rail.

## 1. Chat outage — `GUARDRAIL_HASH_SECRET` (the "errors every time" report)

**Root cause:** commit `aef83dea` (Atrium PR #1062, review round 6) turned a missing `GUARDRAIL_HASH_SECRET` into a **throw when `NODE_ENV=production`**. The guardrails service is constructed lazily on the first chat request (`route → getContentSafetyService → BedrockGuardrailsService`), so the "startup error" surfaced as a 500 on **every** `/api/nexus/chat` request. The matching infra change was never made, so the first deploy carrying that commit bricked chat. CloudWatch (`/ecs/aistudio-dev`) confirms: `GUARDRAIL_HASH_SECRET must be set in production…` on every chat.

**Why it never reproduced locally:** the constructor early-returns into disabled mode when `AWS_REGION` is unset — the failure needs production `NODE_ENV` **+** `AWS_REGION` **+** no secret, which is exactly and only the deployed combination. Reproduced locally with `next build --webpack` + `AWS_REGION=us-east-1`: `{"error":"Failed to process chat request"}` on every message.

**App fix (availability is never traded for the secret):**
- No more throw. Missing secret → loud `log.error` (production) / warn (dev), and `hashValue()` **refuses to hash**, returning a fixed `'redacted'` placeholder instead of falling back to the repo-visible default literal. Zero correlation is possible — **strictly more private** than default-secret HMAC (the original #727 concern) — and chat stays up.

**Infra fix (real hashing after the next deploy):**
- New generated secret `aistudio-{env}-guardrail-hash-secret` (same pattern as `CollabJwtSecret`), threaded through `FrontendStackEcs → EcsService`, injected as the `GUARDRAIL_HASH_SECRET` container secret, added to both IAM secret-read lists. `bunx tsc --noEmit` + `cdk synth` clean.

**Verified:** on the fixed production build (with `AWS_REGION`, no secret) chat streams normally and the ops error logs once; with a secret, ids HMAC-pseudonymize (unit-tested: no-throw, placeholder, 16-hex deterministic HMAC).

## 2. Workspace panel document layout (unchanged from the original PR description)

`DocumentEditor` keys its 288px comment rail off the **viewport** (`hidden w-72 md:block`); inside the 380–720px panel the editor collapsed to ~17% width with character-wrapped text and a clipped toolbar. Fix: `layout="panel"` prop stacks the rail below the editor (page mode untouched), `EditorToolbar` gains `flex-wrap`. Verified at 1512px/1024px: editor ≈95% of panel width, no horizontal overflow; full-page editor byte-for-byte unchanged. Regression guards added (E2E width ratio + unit prop assertion).

## Investigated, not bugs
- **PSD Data MCP 401 "Invalid token format"** — `cognito_passthrough` rejecting a minted E2E token; real sessions authenticate. Degrades gracefully.
- **`/api/notifications/stream` 429** — the 3-connection/user DoS cap under rapid test navigation + StrictMode; self-heals via client backoff.
- **Collab "Connection error" on the local prod server** — `COLLAB_JWT_SECRET` absent locally; correct fail-closed scope (collab only, chat unaffected) and **its ECS wiring already exists** (unlike the guardrail secret).
- **Artifact sandbox-origin notice** — expected local degradation; Edge-runtime compile warnings — pre-existing.

## Known gap (flagged, not in this PR)
The design spec's §1087 "re-prompt path" (chat messages that modify the open artifact call `create_version`) was **never wired into Nexus chat** — PR #1126 shipped the panel as a pure layout sibling with no content tools in the chat toolset. Today chat and panel coexist but chat cannot edit the open document/artifact. That's a feature build (workspace context + content tools in the chat surface), not a regression fix — happy to take it as a follow-up issue.

## Gates
- [x] `bun run typecheck` clean · lint zero warnings on all touched files (incl. clearing the pre-existing `buildConfig` complexity warning)
- [x] Unit: guardrails hash regression (3), workspace panel (5+11), editor tests pass
- [x] E2E on **production build**: 3/3 pass · chat verified streaming across 3 providers
- [x] `infra: tsc --noEmit` + `cdk synth` clean

### DIFF
--- a/components/atrium/DocumentEditor.tsx
+++ b/components/atrium/DocumentEditor.tsx
@@ -52,9 +52,21 @@ export interface DocumentEditorProps {
   idOrSlug: string;
   /** The current user's id, stamped on their edits (green rail). */
   userId: number;
+  /**
+   * Layout context (Epic #1059 §17). `"page"` (default) is the full-width
+   * `/atrium/[id]/edit` page: the 288px comment rail sits BESIDE the editor.
+   * `"panel"` is the narrow Nexus workspace sibling (~380–720px), where a fixed
+   * 288px side rail would collapse the editor to an unreadable sliver — so the
+   * rail stacks BELOW the editor at full width instead.
+   */
+  layout?: "page" | "panel";
 }
 
-export function DocumentEditor({ idOrSlug, userId }: DocumentEditorProps) {
+export function DocumentEditor({
+  idOrSlug,
+  userId,
+  layout = "page",
+}: DocumentEditorProps) {
   // Lazily create the Y.Doc once (a null ref initializer reads clearer than the
   // `undefined as unknown as Y.Doc` cast). The lazy init below runs on the first
   // render before any consumer, so `ydoc` is always a live doc by use.
@@ -205,7 +217,7 @@ export function DocumentEditor({ idOrSlug, userId }: DocumentEditorProps) {
   const { suggesting, count: suggestionCount } = useSuggestionState(editor);
 
   return (
-    <div className="flex flex-col gap-2">
+    <div className={cn("flex flex-col gap-2", layout === "panel" && "p-3")}>
       <EditorToolbar
         status={status}
         canEdit={canEdit}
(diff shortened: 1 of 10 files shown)


krishagel:
## Review round — resolution (commit d02e5677)

**Fixed:**
- **Codex P2 + Gemini — toolbar action group doesn't wrap internally**: correct, and the strongest catch of the round. The `canEdit` group was one flex item, so the outer `flex-wrap` couldn't help below its ~500px min-content width. The group now wraps internally (`flex-wrap justify-end gap-x-2 gap-y-1`). Verified at the panel's 380px minimum (1024px viewport): a DOM sweep of every `button`/`select` in the panel found **zero** controls rendering past the panel edge, with the guard spec still green.
- **Copilot + Gemini — E2E guard robustness**: editor lookup now scoped to the panel element; overflow check allows 1px of subpixel rounding.

**Partially adopted:** Gemini's suggestion to use `locator().boundingBox()` — kept the single `page.evaluate` (one round-trip for panel + editor + overflow, and the elements are already awaited-visible earlier in the test), but took the scoping and tolerance improvements it bundled.

PR #1136, opened 2026-07-06
feat: Nexus workspace chat editing — chat edits the open document/artifact (Atrium §1087)

krishagel:
## Summary

Wires the Atrium design spec's **re-prompt path (§1065/§1087)**: when a document or artifact is open beside the Nexus chat (`?workspace=<id|slug>`, the Epic #1059 §17 panel), the chat can now **read and edit that open object**. Before this, the panel was a pure layout sibling with no content tools in the chat surface — chat and panel coexisted but couldn't interact, so "ask the chat to change the doc" did nothing.

This is the follow-through you asked for after #1131 (which fixed the panel *layout*).

## How it works

**Server** (`lib/nexus/workspace-chat-tools.ts` + `app/api/nexus/chat/route.ts`): a new `workspaceId` request field. When present, the route binds a small AI SDK tool set for **that** object and injects a system-prompt line. Tools are built **server-side** from the resolved id (never the client tool list), and the object is resolved through `contentService` (canView 404-mask → canEdit 403) against the session user. `maxSteps` is raised so the model can read → edit → respond in one turn.

| Tool | When | Effect |
|------|------|--------|
| `read_workspace_content` | always (viewable) | current title/kind/body so edits build on current content |
| `edit_workspace_document` | editable **document** | §28.3-screens markdown → writes to the **live Yjs doc** via the agent bridge (`applyAgentEdit`); appears **immediately** in the panel with agent (purple-rail) attribution; `append`(default)/`replace` |
| `update_workspace_artifact` | editable **artifact** | `contentService.createVersion` (canView/canEdit-gated + §28.3-screened); new version shows in the version dropdown |

A view-only caller gets only the read tool; an unknown/unviewable `?workspace=` yields **no** tools (a bad param never breaks chat). Every tool reuses the existing §11–§15 services — no new content logic, so screening/provenance/visibility/versioning are inherited.

**Client** (`app/(protected)/nexus/page.tsx`): threads the open workspace id to the chat transport via a ref (so opening/closing/switching mid-conversation always sends the current value). **Also fixes a real bug this feature exposed:** switching the model with a workspace open did `window.location.href = '/nexus'`, silently **dropping `?workspace=`** and closing the open document + its tools — it now preserves the param. This is what made the feature appear broken when the user changed models.

## Verification (you asked me to actually test it)

Proven **end-to-end** on the local `:3100` collab server:
- A chat message ("append a paragraph containing `WSMARK…`") drove `read_workspace_content` → `edit_workspace_document`, and the text appeared **live in the open editor with the purple agent rail** (screenshot: two chat-driven markers, both purple/agent-attributed).
- An artifact chat edit created **version 2** (`author_actor` recorded, body = the requested code), visible in the version dropdown.

Committed tests:
- **Unit** `tests/unit/lib/nexus/workspace-chat-tools.test.ts` (9): gating (unresolved requester / unviewable → null), read-only vs editable tool sets, kind-specific tool selection, document screen-then-bridge, screening-refusal blocks the write, artifact `createVersion` wiring, read-result shape.
- **E2E** `tests/e2e/nexus-workspace-chat-tools.spec.ts`: the chat request carries `workspaceId` **and** `?workspace=` survives a model change — deterministic (intercepts the request; no live-LLM dependency in CI).
- Existing `nexus-workspace-panel` + `nexus-chat-tools` E2E still pass (no regression to the fragile conversation tree).

## Gates
- [x] `bun run typecheck` clean
- [x] `bun run lint` — zero warnings on the new module + route; `page.tsx` pre-existing warnings unchanged (7→7; they're on the 192/270-line Nexus functions I deliberately didn't refactor per CLAUDE.md rule #6)
- [x] Unit (41 in the affected suites) + 3 E2E pass on the authed :3100 server

No new DB migrations, no infra changes.

### DIFF
--- a/app/(protected)/nexus/page.tsx
+++ b/app/(protected)/nexus/page.tsx
@@ -68,6 +68,8 @@ interface ConversationRuntimeProviderProps {
   enabledTools: string[]
   enabledConnectors: string[]
   skillId?: string
+  /** The open workspace object id/slug (`?workspace=`); binds §1087 content tools server-side. */
+  workspaceId?: string
   attachmentAdapter: AttachmentAdapter
   voiceAdapter?: RealtimeVoiceAdapter
   initialMessages?: UIMessage[]
@@ -88,6 +90,7 @@ function ConversationRuntimeProvider({
   enabledTools,
   enabledConnectors,
   skillId,
+  workspaceId,
   attachmentAdapter,
   voiceAdapter,
   initialMessages = [],
@@ -101,6 +104,13 @@ function ConversationRuntimeProvider({
   // causing duplicate message rendering. (Issue #868)
   const conversationIdRef = useRef(conversationId)
   conversationIdRef.current = conversationId
+
+  // Track the open workspace id via ref so the transport body always sends the
+  // CURRENT value: the runtime is created once (stable), but the user can open /
+  // close / switch the workspace panel mid-conversation (§1087). Reading a ref
+  // keeps the sent workspaceId fresh without recreating the transport.
+  const workspaceIdRef = useRef(workspaceId)
+  workspaceIdRef.current = workspaceId
 
   // Track session status via ref for use inside customFetch without adding to deps.
   // useSession is safe here — SessionProvider wraps this component tree.
@@ -323,6 +333,8 @@ function ConversationRuntimeProvider({
           enabledConnectors,
           // Bind the session to a skill so the server enforces its allowed-tools pin (#925).
           skillId,
+          // Bind the open workspace object so the server offers §1087 read/edit tools.
+          workspaceId: workspaceIdRef.current || undefined,
           conversationId: conversationIdRef.current || undefined
         }
       }
@@ -353,6 +365,8 @@ interface NexusRuntimeWrapperProps {
   enabledTools: string[]
   enabledConnectors: string[]
   skillId?: string
+  /** Open workspace object id/slug (`?workspace=`); passed to the runtime for §1087 tools. */
+  workspaceId?: string
   attachmentAdapter: AttachmentAdapter
   voiceAvailable: boolean
   voiceUnavailableReason?: string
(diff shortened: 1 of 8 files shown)


krishagel:
## Review round — resolution (commit 10826a0d)

Thanks — three real issues here, all fixed.

**Fixed:**
- **P1 (Gemini ×3 + Codex) — artifact writes skipped §28.3 screening**: correct and the most important catch. The tool runs under a `kind:"user"` requester and `createVersion`'s internal screening only covers *agent* requesters, so model-generated artifact code was persisted unscreened. Now `buildArtifactUpdateTool` takes `requestId` and calls `screenAgentContent` before `createVersion`, failing closed — mirroring the document path. Added screen-then-version + screening-refusal unit tests.
- **P1 (Codex) — `read_workspace_content` returned only `bodyInline`**: real. Documents keep text in Yjs (bodyInline null) and large artifacts store source at `bodyLocation`. Documents now read the `atrium_doc_state.markdown` projection; when content is genuinely unavailable the tool returns `bodyUnavailable: true` (never a false empty) so the model won't rewrite from nothing. (I tried decoding the live Yjs state directly, but y-prosemirror's serializer needs the editor schema and throws server-side, so the projection + honest `bodyUnavailable` is the robust source.)
- **P2 (Codex) — skill `allowed-tools` pin bypassed**: real. Workspace tools were merged after skill binding unfiltered. `applySkillSessionBinding` now returns the pin and `filterWorkspaceToolsBySkillPin` drops workspace tools the skill didn't allow.
- **P2 (Codex) — provider-native tools dropped**: real (and pre-existing for connectors). `getOrCreateTools` now **merges** the model's provider-native tools under the pre-merged set instead of replacing; the route passes `enabledTools` alongside `tools`. Benefits connector sessions too.
- **Copilot** — title `JSON.stringify`-escaped before the system prompt (injection guard); read `body` returns `null` (not `""`) when absent.

**Verified live on the :3100 collab server**: document read+edit and artifact screen+version still work, plain chat + provider-tool merge intact, committed E2E green; 12 unit tests.

(Note: the required `claude-review` check is failing repo-wide on a transient `Failed to parse JSON` / empty `ANTHROPIC_API_KEY` — a CI-config issue, not this PR.)

PR #1137, opened 2026-07-06
fix: two deployed-dev blockers for Nexus workspace tool calls (guardrail-profile IAM grant + collab WS bundle)

krishagel:
## Why

After deploying the §1087 chat-edits-workspace feature (#1136), the deployed **dev** environment showed two runtime failures the local Playwright harness could not exercise (Bedrock screening is disabled without creds; the WS bundle only exists in the built image):

- Chat replied, but every **workspace tool call** returned `Content could not be safety-screened right now. Please retry shortly.` ("temporary issue with the safety screening system").
- The workspace **panel stayed stuck on "Connecting…"**.

Both are deploy/packaging-level, not app-logic, and both were found in CloudWatch `/ecs/aistudio-dev`.

## Fix 1 — grant `bedrock:ApplyGuardrail` on the guardrail-**profile** ARN

CloudWatch:
```
User: ...EcsServiceTaskRole... is not authorized to perform: bedrock:ApplyGuardrail
on resource: arn:aws:bedrock:us-east-1:...:guardrail-profile/us.guardrail.v1:0
```
The guardrail uses **cross-region inference**, so `ApplyGuardrail` authorizes against **both** the guardrail ARN **and** the region-qualified `guardrail-profile/*` ARN. The task role was granted only the guardrail ARN → 100% AccessDenied on the profile. That made `ContentSafetyService` fail and, critically, made **Atrium §28.3 agent-write screening fail closed** — every workspace/agent content edit was rejected. Same class as the agent-router fix in #1093.

`infra/lib/constructs/ecs-service.ts`: add `guardrail-profile/*` to both the specific-ARN and wildcard branches of the `ApplyGuardrail`/`GetGuardrail` statement.

## Fix 2 — build the voice + collab WS bundles in ONE Docker layer

CloudWatch (same image):
```
[atrium-collab] Collab WebSocket disabled: handler not found at ./collab-handler-bundle.cjs
[voice-server] WebSocket handler registered       ← sibling bundle loaded fine
```
Both bundles are esbuild-produced into `.next/standalone/` and picked up by the same `COPY .next/standalone ./`. Voice loaded, collab did not — the collab bundle simply wasn't in the built image. They were **two separate `RUN` layers**, and the second layer's output was not surviving into the standalone `COPY` under BuildKit (CDK `fromAsset` build). Without the collab handler the workspace collab WS never starts → panel "Connecting…" forever and agent doc edits can't push.

`Dockerfile`: build **both** bundles in a single `RUN` layer and `test -f` **assert** both exist, so a missing bundle is a loud build failure instead of a silent runtime one.

## Deploy note

Takes effect on the next **FrontendStack** deploy: the IAM grant updates the task role, and CDK `fromAsset` rebuilds the image (picking up the collab bundle). The canonical full deploy covers both.

## Testing
- `infra` `tsc --noEmit` clean.
- Verified locally both bundle scripts emit their `.cjs` into `.next/standalone/` with only Node built-in externals (`voice: OK / collab: OK`), then removed the local artifacts.
- Screening AccessDenied and the collab-disabled boot line are deploy-only; they were diagnosed from CloudWatch and cannot be reproduced by the dev-server harness (guardrails self-disable without AWS creds; the WS bundle exists only in the built image).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,14 +54,23 @@ ENV RDS_SECRET_ARN=${RDS_SECRET_ARN}
 RUN --mount=type=cache,target=/app/.next/cache \
     node_modules/.bin/next build --webpack
 
-# Bundle voice WebSocket handler as a self-contained CJS artifact for voice-server.js.
-# The script fails the build if any non-builtin runtime externals leak into the bundle.
-# Writing into .next/standalone ensures the runner-stage COPY picks it up automatically.
-RUN node scripts/build-voice-ws-handler.mjs
-
-# Bundle the Atrium collaboration WebSocket handler the same way (#1051), so
-# voice-server.js can load ./collab-handler-bundle.cjs at runtime.
-RUN node scripts/build-collab-ws-handler.mjs
+# Bundle the voice AND Atrium-collab WebSocket handlers as self-contained CJS
+# artifacts for voice-server.js. Each script fails the build if a non-builtin
+# runtime external leaks in. Writing into .next/standalone ensures the runner-stage
+# COPY picks them up.
+#
+# BUILT IN ONE LAYER (2026-07-06): they were previously two separate RUN steps.
+# In the deployed image the voice bundle was present but the collab bundle was
+# MISSING ("Cannot find module ./collab-handler-bundle.cjs" → collab disabled →
+# workspace panel stuck "Connecting…" and agent doc edits fail) — the second
+# RUN layer's output was not surviving into the standalone COPY under BuildKit.
+# Producing both in a single layer (the one that demonstrably IS copied) and
+# asserting both files exist makes a missing bundle a LOUD build failure instead
+# of a silent runtime one.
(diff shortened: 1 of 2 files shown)


krishagel:
Addressed the guardrail-profile ARN review in 62daf91a:

- **Copilot (adopted):** scoped `guardrail-profile/*` → the exact `guardrail-profile/us.guardrail.v1:0` in both branches, matching `guardrails-stack.ts:287` and `agent-platform-stack.ts:865/1079` (the proven PR #1093 agent-router grant). Least-privilege + codebase-consistent.
- **Gemini (declined — false positive):** the high-priority suggestion to wildcard the region (`arn:aws:bedrock:*:...`) contradicts the working precedent. `ApplyGuardrail` authorizes against the profile ARN in the **calling** region; ECS runs in us-east-1 and calls bedrock-runtime in us-east-1, so the region-qualified stack-region ARN is correct and is exactly what the already-deployed, working agent-router/guardrails grants use.

Issue #1138, opened 2026-07-07
AI Agent multi-failure incident: attachments never reach agent, Drive scope 404s, mid-task abort from 100k context cap, error turns leak internal narration, guardrail IAM AccessDenied on every message


krishagel:
# PSD AI Agent — multi-failure incident, 2026-07-06 evening (Google Chat DM with hagelk)

Five distinct defects surfaced in one conversation. Each is root-caused below with code references and CloudWatch evidence from `/aws/lambda/psd-agent-router-dev` and `/aws/bedrock-agentcore/runtimes/psd_agent_dev-gdrXYv6E54-DEFAULT` (dev account 390844780692, us-east-1). All timestamps UTC.

## Incident timeline (from screenshots + logs)

| UTC | Event |
|---|---|
| 22:48:49 | "Do you have the ability to read this?" + **Google Doc chip** → agent: "I don't see any attachment, link, or file" |
| 22:49:35 | User pastes the Drive URL → agent: Drive returns **"File not found"**, asks user to share with `hagelk@psd401.net` |
| 22:51:21 | User uploads the **PDF directly in Chat** → agent: "I don't see a file attached on my end" |
| 23:34:46 | Plaud transcript → 2 Google Docs → share → post-to-POAD task starts |
| 23:41:12 | Turn dies after 386 s / **89 tool calls / 3,895,032 total tokens**; user receives leaked internal narration: *"Now let's find recordings from 7/1 in Plaud.Good, done. Let's read the file.Now getting the action-items breakdown for the todo list doc."* Router log: `Agent returned an error turn — errorClass: OpenClawChatError` |
| 23:51:06 | "Were you still working?" → **"I processed your message but had no response."** (`errorClass: EmptyAgentResponse`) |
| every message | Router logs `Guardrail invocation failed — AccessDenied` (see F5) |

---

## F1 — Chat attachments and Drive chips never reach the agent

**Root cause:** `infra/lambdas/agent-router/index.ts` forwards only text. The `GoogleChatEvent` interface (line ~203) models `text` / `argumentText` / `sender` / `thread` / `slashCommand` only. Google Chat's message resource also carries:

- `attachment[]` — files uploaded in Chat (the PDF), with `attachmentDataRef` / `driveDataRef` + `contentType`
- `annotations[]` — Drive chips / rich links (the Doc chip), with `richLinkMetadata.driveLinkData.driveDataRef.driveFileId`

[...]

PR #1140, opened 2026-07-07
fix: Atrium §28.3 agent-write screening fails OPEN on degraded guardrails

krishagel:
## Why

On deployed dev, every Nexus **workspace tool call** (chat editing the open document/artifact) returned:

> Content could not be safety-screened right now. Please retry shortly.

The upstream trigger — an `ApplyGuardrail` AccessDenied on the cross-region guardrail-**profile** ARN — is being fixed separately (#1138). But the **design defect is on the content path**: `screenAgentContent` treated a *degraded* guardrail evaluation (any AWS error / timeout / throttle / IAM denial) as a hard **block** — it failed **closed**. Guardrails on this path are telemetry-only and must never block a write on their own unavailability.

## Change

`lib/content/agent-screening.ts`: on `safety.degraded`, log the skipped evaluation as telemetry and **allow** the content (`{allowed:true}`) instead of returning a degraded rejection. Only a **positive** guardrails detection (`allowed:false`) still refuses content.

Consumers inherit the core verdict:
- `app/api/content/[id]/agent-bridge/route.ts` — dropped the now-dead 503 branch; a non-allowed verdict is always a 422 block.
- `lib/nexus/workspace-chat-tools.ts` — comment-only (logic already keys off `!allowed`).
- Removed the now-unused `AGENT_SCREEN_DEGRADED_MESSAGE` and the `"degraded"` verdict variant; `screenAgentBodyForWrite` no longer throws "Safety screening unavailable".

## Verification
- **36 unit tests** pass — `atrium-agent-screening`, `-wiring`, and `workspace-chat-tools`; degraded cases now assert fail-OPEN (allow / proceed to transaction).
- **Authenticated Playwright on :3100** — `nexus-workspace-panel` (panel + live collab editor + close-preserves-chat) and `nexus-workspace-chat-tools` (workspaceId in request, survives model change) both pass.
- **Real (non-mocked) degraded chain** — guardrails ENABLED with a bad id + no creds → the real `ApplyGuardrail` SDK call errors → `degraded:true` → `screenAgentContent` logs "failing open" and returns `{allowed:true}`. This is the exact deployed condition, now allowing instead of blocking.
- **Not covered locally:** a full real-LLM edit (no LLM provider keys in this env). The model→tool hop is unchanged from #1136; only the screening verdict changed.

## Scope
Content-path screening only. Does **not** touch the guardrail IAM grants (#1138 owns those).

### DIFF
--- a/app/api/content/[id]/agent-bridge/route.ts
+++ b/app/api/content/[id]/agent-bridge/route.ts
@@ -198,10 +198,11 @@ async function loadEditableObject(
 }
 
 /** Screen agent markdown via the shared §28.3 core (`lib/content/agent-screening`
- * — guardrails fail-closed + PII telemetry, logging included). `requestId` threads
- * this request's correlation onto the core's blocked/degraded log lines. This
- * wrapper only maps the verdict to HTTP: 422 for blocked content, 503 for a
- * degraded (unavailable) screening evaluation. Returns null to proceed. */
+ * — guardrails fail-OPEN + PII telemetry, logging included). `requestId` threads
+ * this request's correlation onto the core's blocked log lines. This wrapper maps
+ * the verdict to HTTP: 422 for a positive guardrails detection. A degraded
+ * (unavailable) evaluation fails open in the core and never reaches here — the
+ * write proceeds. Returns null to proceed. */
 async function screenAgentMarkdown(
   markdown: string,
   objectId: string,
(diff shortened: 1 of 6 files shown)


krishagel:
🤖 **pr-fix routine — round 1**

Addressed the two doc nits from the review above in 7001ce4:
- `lib/content/agent-screening.ts`: dropped the dead "422/503" reference (the bridge's 503 branch was removed by this PR; only 422 remains).
- `lib/safety/types.ts`: updated the `degraded` field's JSDoc, which still claimed the Atrium agent bridge fails CLOSED on a degraded evaluation — this PR's whole point is that it no longer does.

**Not actioned: the fail-open security-regression concern.** The reviewer is asking whether it's acceptable to let unscreened agent-authored content persist into K-12 documents during any guardrails outage (not just the #1138 IAM misconfig), and wants explicit product/compliance sign-off given the FERPA/COPPA/CIPA implications. That's a policy tradeoff, not a code defect — this autonomous routine isn't positioned to decide it, and reverting or rescoping the behavior unilaterally would be exactly the kind of unreviewed judgment call it shouldn't make. @krishagel — worth an explicit decision here: accept the tradeoff as-is (and update the stale doc/JSDoc language accordingly, now done), or take one of the reviewer's alternatives (keep fail-closed here and rely on #1138 fixing the IAM grant, or bound the fail-open window with alerting).

Labeling `pr-fix-stuck` so the routine doesn't re-loop on this — remove the label once the tradeoff is decided (and I'll pick up any resulting follow-up work, including doc updates, next run).




---

PR #1184, opened 2026-07-11
fix(agent-image): mid-turn compaction guard, bedrock memory plugin, venv-absolute python

krishagel:
## Summary
Fixes the three production bugs found in tonight's AgentCore runtime logs (2026-07-10 evening session, `/aws/bedrock-agentcore/runtimes/psd_agent_dev-*`):

1. **Context overflow killing long turns** — `compactionAttempts=0` at ~180k tokens: OpenClaw 2026.6.11 only auto-compacts at turn start; the mid-tool-loop guard is opt-in and was off. Enables `agents.defaults.compaction.midTurnPrecheck.enabled` (verified against the pinned tag's docs — truncates/compacts + retries instead of hard-failing). Root cause of both '⚠️ I couldn't finish that' aborts and the 'busy processing another request' session-lock replies.
2. **Memory embedding sync dead** — 'Unknown memory embedding provider: bedrock' every cycle: the `bedrock` adapter is registered by `@openclaw/amazon-bedrock-provider`, and the published base image ships zero bundled extensions. Vendors the plugin (pinned `2026.6.11`, self-contained tarball: bundled @aws-sdk deps + shrinkwrap, manifest-existence asserted) under `/opt/openclaw-plugins` — image-owned, NOT the runtime-persistent `~/.openclaw` — folded into the existing consolidated RUN (54-layer Firecracker ceiling, no new layer), loaded via `plugins.load.paths`, model discovery OFF (embedding contract only; Mantle remains the sole chat path). Auth (`AWS_BEARER_TOKEN_BEDROCK`, `AWS_REGION`) already wired.
3. **TTS/boto3 spawn-context failure** — plain `python3` in SKILL.mds resolves to the boto3-less system python when spawned from node child processes (hit live tonight by the user's morning-brief skill). Hardens `psd-tts`, `psd-last30days` (S3-HTML mode also needs boto3), and `psd-pdf-to-markdown` to absolute `/opt/agentcore-venv/bin/python3`.

## Verification
- openclaw.json parses; every config key verified against openclaw `v2026.6.11` source/docs (`compaction.midTurnPrecheck`, `plugins.load.paths`, plugin manifest `enabledByDefault`/`memoryEmbeddingProviders`).
- `npm pack` smoke-tested locally: tarball name, `package/` layout, `openclaw.plugin.json` (id `amazon-bedrock`, embeddings `[bedrock]`), bundled `node_modules/@aws-sdk` all confirmed.
- The probe-gated image build (boot probe + canary turn) runs in parallel as the functional gate — CI does not build Docker images.

Post-deploy checks for tomorrow's 6 AM 'Morning Dispatch' run: no more `memory sync failed (watch)` spam, no `Context overflow ... compactionAttempts=0`, TTS produces the podcast MP3.

https://claude.ai/code/session_019dVBkHTmcbtyFHRDLJi1Sy

### DIFF
--- a/infra/agent-image/Dockerfile
+++ b/infra/agent-image/Dockerfile
@@ -374,6 +374,19 @@ COPY skills /opt/psd-skills
 # "Failed to mount overlay: No such file or directory" (agent outage 2026-07-02,
 # caused by adding the gws-wrapper COPY at 53). Absolute `cd` paths so the chain
 # is independent of prior CWD. Keep new skills in this RUN, not new RUNs.
+#
+# The tail of this RUN also vendors the @openclaw/amazon-bedrock-provider
+# plugin: it registers the "bedrock" memory-embedding adapter that
+# openclaw.json's memorySearch.provider names. The published OpenClaw base
+# image ships with NO bundled extensions (they're an opt-in OPENCLAW_EXTENSIONS
+# build arg of the upstream Dockerfile), so memory sync failed every cycle with
+# "Unknown memory embedding provider: bedrock" (2026-07-10). It must live under
+# /opt (image-owned): ~/.openclaw is runtime-persistent storage that shadows
+# anything written there at build time. Loaded via openclaw.json
+# plugins.load.paths. Version-pinned to the same OpenClaw release this image
+# pins (2026.6.11); npm verifies the registry integrity hash on pack. The
+# tarball is self-contained (bundleDependencies: @aws-sdk clients ship inside
+# it, with an npm-shrinkwrap.json), so no npm install runs after extraction.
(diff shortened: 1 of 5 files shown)


krishagel:
🤖 gemini's finding applied in b3efba30 — `allowed-tools` now permits ONLY `Bash(/opt/agentcore-venv/bin/python3:*)` in all three SKILL.mds (plain `python3` removed).

Probe-gate result: the image built from this branch **passed boot probe + canary** (tag `2026-07-10-runtime-hardening`) — the vendored amazon-bedrock plugin loads and the new openclaw.json keys are accepted at runtime. A final rebuild happens post-merge to pick up the allowed-tools tweak.

Issue #488, opened 2025-11-25
bug: Nexus conversation history not loading when navigating to existing conversations


krishagel:
## Description

When clicking on a conversation in the Nexus sidebar or navigating directly to a URL with `?id=<conversation-id>`, the conversation history never loads. The page shows "Loading conversation..." indefinitely, and critically, **no network request is made** to fetch the conversation data from `/api/nexus/conversations/{id}/messages`.

## Steps to Reproduce

1. Open Nexus and create a conversation with some messages
2. Click on that conversation in the sidebar (or navigate directly via URL: `http://localhost:3000/nexus?id=8c1ccbf9-ed9b-4390-82f7-2e970226ffb4`)
3. Observe: Page shows "Loading conversation..." but messages never appear
4. Check Network tab: **No API call is made to `/api/nexus/conversations/*/messages`**
5. Check Console: No errors related to conversation loading

## Expected Behavior

When navigating to an existing conversation:
- API call should be made to `/api/nexus/conversations/{id}/messages`
- Messages should be loaded and displayed in the chat
- User can continue the conversation

## Actual Behavior

- "Loading conversation..." spinner shows indefinitely
- No network call to fetch messages
- `historyAdapter.load()` is never invoked
- Chat remains empty despite having a valid `conversationId` in the URL

### Breaking Change

**Commit:** `99f569f` (Oct 4, 2025) - "feat: migrate Nexus Chat from polling to native SSE streaming (#308)"

**What Changed:**
```typescript
// BEFORE (working) - used useLocalRuntime
const runtime = useLocalRuntime(streamingAdapter, {
  adapters: {
    history: historyAdapter,  // ← load() called automatically on mount
  }
})

// AFTER (broken) - uses useChatRuntime  
const runtime = useChatRuntime({
  transport: new AssistantChatTransport({...}),
  adapters: {
    history: historyAdapter,  // ← load() is NEVER called
  }
})
```

### Why It Broke

1. **`useLocalRuntime` behavior:** Automatically invokes `historyAdapter.load()` on mount/conversation change
2. **`useChatRuntime` behavior:** Does NOT automatically invoke `historyAdapter.load()` - requires manual import via `runtime.threads.main.import()`
3. **Evidence:** [assistant-ui GitHub Issue #2365](https://github.com/assistant-ui/assistant-ui/issues/2365) - "Unable to preload existing messages to useChatRuntime thread"

### Why Prompt Library Still Works

[...]

Issue #480, opened 2025-11-18
fix: Conversation history not loading in Nexus on AWS (dev/prod), works locally


krishagel:
## Description
Conversation loading shows infinite spinner in Nexus when loading previous conversations on AWS-hosted dev and prod environments. Same conversations load successfully on localhost:3000.

## Steps to Reproduce
1. Navigate to Nexus on AWS dev or prod environment
2. Attempt to load an existing conversation (e.g., ID: dbed33d4-e368-4ce7-b2ae-79cd09cf384a)
3. Observe infinite loading spinner
4. Compare with localhost:3000 - conversation loads successfully

## Expected Behavior
Conversations should load on AWS environments with the same speed and reliability as localhost

## Actual Behavior
Conversations fail to load on AWS (infinite spinner), but load successfully on localhost

## Environment Details
- **Failing**: AWS-hosted dev and prod environments
- **Working**: Local development (localhost:3000)
- **User**: Kris Hagel
- **Conversation ID**: dbed33d4-e368-4ce7-b2ae-79cd09cf384a
- **Browser**: Chrome 142.0.0.0 on macOS 10.15.7
- **Timestamp**: 2025-11-18T17:42:28.145Z
- **Network**: 4g connection, 4.1 Mbps downlink
- **JS Heap**: 75 MB used / 4096 MB limit

## Root Cause Analysis
Based on code analysis, the issue appears to be related to the conversation loading endpoint `/api/nexus/conversations/[id]/messages` which is defined in `app/api/nexus/conversations/[id]/messages/route.ts:30`.

### Potential Issues Identified:
1. **Database connectivity**: The endpoint uses RDS Data API via `executeSQL` function from `lib/streaming/nexus/db-helpers.ts:20`
2. **Environment configuration**: RDS_RESOURCE_ARN and RDS_SECRET_ARN are passed to ECS tasks (infra/lib/constructs/ecs-service.ts:421-422)
3. **Authentication**: Uses session-based auth which may behave differently in AWS vs localhost
4. **Request timeout**: API requests may be timing out on AWS due to network latency

### API Endpoint Details:
- **Route**: `GET /api/nexus/conversations/[id]/messages`
- **File**: `app/api/nexus/conversations/[id]/messages/route.ts`
- **Database Query**: Fetches messages from `nexus_messages` table ordered by created_at
- **Authentication**: Requires valid session and verifies user ownership of conversation
- **Pagination**: Supports limit (max 1000) and offset parameters

[...]

PR #1200, opened 2026-07-13
feat(atrium): hard delete for content objects (owner/admin, unpublish-first)

krishagel:
## Summary

Atrium (Epic #1059) shipped with no way to remove content — only status archival. This adds a real, safety-gated **hard delete** for content objects across every surface (service, REST v1, psd-atrium skill, Nexus workspace chat, and the Meridian UI), so the agent and users can clean up junk content.

## Guard semantics (all must hold; ordered so existence is never revealed)

- **404** existence-mask (`canView`) BEFORE any permission signal;
- **403** unless the requester is the object's **owner** or an **admin** (agents also need the new `content:delete` scope) — a dedicated `assertCanDelete` predicate, deliberately not `canEdit`, so a future widening of edit can never hand out delete;
- **409** while the object is **live at any publish destination** — delete NEVER auto-retracts; the caller must unpublish first. Checked as a fast pre-flight AND authoritatively on the FOR-UPDATE-locked row inside the delete transaction (TOCTOU-safe).

Any kind/status otherwise (draft / archived / private / internal) is deletable — the guards are the protection, not the lifecycle status.

## Cascade (one transaction; external cleanup ordered so failure can't orphan DB state)

- Retrieval-index removal runs via the sanctioned `retrievalService.removeFromIndex` **before** the tx (it reads `content_index_links` to find the SHARED `repository_item`; the object-delete cascade would otherwise erase that link and orphan the shared rows).
- In the tx: write a `delete` audit row capturing the removed object's title/kind/owner/versionsDeleted (the object_id becomes a dangling UUID after — the only durable record of what was removed), delete the object's intranet **nav entry** (`navigation_items.content_object_id` is `NO ACTION`, and unpublish only soft-hides it), then delete the object row — cascading `content_versions`, `content_publications`, `atrium_doc_state`, `atrium_doc_comments`, `content_embed_links`, `content_visibility_grants`, `content_publish_requests`, `content_index_links`. `content_audit_logs` has no FK to the object, so the trail SURVIVES.
- After commit (best-effort, logged): S3 body-tree delete (`atrium/objects/{id}/`) and a `content.deleted` event. Orphaned S3 keys are acceptable.

## Surfaces

- `contentService.delete` + `canDelete`/`assertCanDelete`.
- `DELETE /api/v1/content/{id}` — new `content:delete` scope (API_SCOPES + `ROLE_SCOPES.staff`; administrator via ALL_SCOPES; added to the key-bootstrap `KEY_SCOPES` so the scope-tether test stays green and the deployed agent key self-heals on the next deploy).
- psd-atrium skill: `delete` subcommand (owner / live-publication refusals relayed verbatim); Phase-1-no-delete claims removed.
- Nexus workspace chat: `delete_workspace_content` tool (session user, owner/admin, same service guards).
- UI: Meridian "Delete permanently" affordance in ContentSettings (owner/admin server-gated; confirm-permanence dialog; disabled with an explanation while a publication is live). Reader/editor routes 404 after deletion.
- Collab hardening: `getOrCreateDoc` now fails closed for a deleted object (a pre-minted collab token can no longer open a just-deleted doc's room within its TTL).

## Verification (all green before opening)

- Build/lint (zero-warning): ✅   Typecheck: ✅ (only pre-existing `.next/types` errors)
- Unit: ✅ 689 passed (full Atrium + Nexus + scopes suites), incl. new coverage for every guard (404-before-403, 403 owner/admin, 409 live pre-flight + in-tx TOCTOU, admin override, audit-with-details, removeFromIndex-before-tx ordering, nav + object cascade delete, S3 best-effort), the `canDelete` predicate, the embed-of-deleted mask, and the workspace `delete_workspace_content` tool.
- E2E: ✅ authed `atrium-delete.functional` (all 3) — create→delete→404, published→409→unpublish→delete, and the UI "Delete permanently" flow (open editor settings → confirm → back to library → 404).

## Evidence

The UI delete flow is exercised end-to-end by the authenticated E2E spec `tests/e2e/atrium-delete.functional.spec.ts` (test 3: open the editor content-settings dialog → click "Delete permanently" → confirm → land back on the library → the object 404s), which passes against the real service + DB on the :3100 host server. No static screenshots were captured for this change.

## Deploy-gated

- Migration 105 (`content_audit_logs.details`) applies on the next DB deploy (applied to local Postgres for the E2E run).
- The deployed psd-atrium agent key re-mints to the new `content:delete` scope set on the next AgentPlatformStack deploy (scope-drift re-mint).
- The agent image must be rebuilt/pushed for the skill's `delete` subcommand to reach the running agent.

### DIFF
--- a/actions/db/atrium/delete-content.ts
+++ b/actions/db/atrium/delete-content.ts
@@ -1,0 +1,82 @@
+"use server"
+
+/**
+ * Atrium delete-content server action (Atrium hard delete)
+ *
+ * The in-app (logged-in human) surface for PERMANENTLY deleting a content
+ * object — the Meridian "Delete" affordance in the editor's content-settings
+ * dialog. A thin wrapper over `contentService.delete`, which owns the guards:
+ * existence-masking (404 before 403), the owner/admin `assertCanDelete` gate, and
+ * the live-publication refusal (409 — delete never auto-unpublishes). It also
+ * writes the transactional `delete` audit row and cleans up the cascade + S3.
+ *
+ * Gate ordering mirrors `update-content.ts`: requester FIRST (an unauthenticated
+ * caller gets 401 "please log in", not 403), then the `atrium-content`
+ * feature-capability gate, then the service. `surface: "ui"` tags the audit row
+ * as an in-app human action.
+ *
+ * See docs/features/atrium-design-spec.md §11.2.
+ */
+
+import {
+  createLogger,
+  generateRequestId,
+  startTimer,
+  sanitizeForLogging,
+} from "@/lib/logger";
+import { createSuccess, handleError, ErrorFactories } from "@/lib/error-utils";
+import { contentService } from "@/lib/content";
+import type { DeletedContentSummary } from "@/lib/content";
+import type { ActionState } from "@/types";
+import { hasCapabilityAccess } from "@/utils/roles";
+import { getServerSession } from "@/lib/auth/server-session";
+import { getUserRequester } from "./requester";
+
+export async function deleteContentAction(
+  // Accepts a UUID OR a slug — `contentService.delete` resolves via
+  // `loadByIdOrSlug`, matching the other slug-tolerant actions.
+  idOrSlug: string
+): Promise<ActionState<DeletedContentSummary>> {
+  const requestId = generateRequestId();
+  const timer = startTimer("deleteContentAction");
+  const log = createLogger({ requestId, action: "deleteContentAction" });
+
+  try {
+    log.info("Action started: delete content", {
+      idOrSlug: sanitizeForLogging(idOrSlug),
+    });
+
+    if (!idOrSlug) {
+      throw ErrorFactories.missingRequiredField("idOrSlug");
+    }
+
+    // Session resolved ONCE, threaded through the requester build + capability
+    // check (see update-content.ts for the same-session rationale). Requester
+    // FIRST so an unauthenticated caller gets a 401, not a 403.
+    const session = await getServerSession();
+    const requester = await getUserRequester(requestId, session);
+    if (!(await hasCapabilityAccess("atrium-content", session!.sub))) {
+      throw ErrorFactories.authzToolAccessDenied("atrium-content");
+    }
+
+    // The service enforces existence-masking (404 before 403), the owner/admin
+    // delete gate, and the live-publication refusal (409). `ui` = in-app human.
+    const result = await contentService.delete(requester, idOrSlug, {
+      surface: "ui",
+    });
+
+    timer({ status: "success" });
+    log.info("Content deleted", {
+      objectId: result.id,
+      versionsDeleted: result.versionsDeleted,
+    });
+    return createSuccess(result, "Content deleted");
+  } catch (error) {
+    timer({ status: "error" });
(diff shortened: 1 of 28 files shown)


krishagel:
## Review findings — addressed in `f9bb596a`

Thanks for the thorough reviews. Summary of how each finding was handled:

### Fixed (code changes in `f9bb596a`)
- **Index removal not rolled back on the in-tx TOCTOU abort** (Codex P2 / Claude #1). `removeFromIndex` must run *before* the delete tx (it needs the not-yet-cascaded `content_index_links` to find the shared `repository_item` via the sanctioned inverse; running it post-commit would leave the shared rows orphaned). So instead of reordering, `contentService.delete` now **re-indexes on abort**: the `ConflictError` catch fires `retrievalService.indexObject` (best-effort, fire-and-forget), so a refused delete never leaves still-live content out of retrieval. The `atrium-content-delete` TOCTOU test now asserts both `removeFromIndex` **and** `indexObject` ran.
- **`delete_workspace_content` bound on `canEdit`, not `canDelete`** (Claude minor). Now bound on `canDelete` so the LLM affordance honors the same owner/admin-only decoupling `helpers.ts` documents (a new workspace-chat-tools test pins `canEdit true + canDelete false → tool not bound`). The service already re-checked `assertCanDelete` per call, so there was no bypass — this is defense-in-depth alignment.

### Reasoned replies (no code change)
- **Active in-memory collab room not evicted on hard delete** (Codex P1 / Gemini HIGH / Claude #2). This is the deliberately-scoped-out edge: a *complete* fix is cross-instance room eviction on `content.deleted` (SNS/Redis fan-out) = session-kill plumbing, out of scope for this PR by design. What *is* enforced: (1) the collab **mint route** re-runs `canView`/`loadByIdOrSlug` and 404s every **new** connection (needs a fresh token); (2) `getOrCreateDoc` now fails closed on cold load; (3) an already-open session's writes can't persist — `atrium_doc_state.object_id` is `ON DELETE CASCADE`, so post-delete `saveDocState` hits the FK and is dropped. So the narrow residual (a co-editor already connected via a visibility grant when an owner/admin deletes) can read/edit in-memory until the room drains, but never persists. Documented in the commit + code comment; a proper cross-instance eviction is a reasonable follow-up.
- **Auth-before-authz ordering** (Gemini MEDIUM, `delete-content.ts:58`). False positive: `getUserRequester` (line 57) throws `authNoSession()` → **401** when the session is null, *before* the `hasCapabilityAccess` check on line 58. So authentication precedes authorization, and the `session!.sub` assertion is safe by that throw-on-null contract — identical to the established `update-content.ts` pattern.
- **Nexus chat tools don't check the `atrium-content` capability** (Claude minor). Pre-existing cross-cutting gap for *all* workspace-chat tools (edit/publish too), not specific to delete. The authoritative gate is the service's `assertCanDelete` (owner/admin). A capability gate for the whole workspace-chat surface is a sensible follow-up rather than a delete-only change here.
- **CodeQL `js/user-controlled-bypass` on `delete-content.ts:49`** — dismissed as a false positive (required-field presence guard, not an authz gate; real authz runs after). Sibling precedent: #488/#487 (`update-content.ts`), #480/#479 (`set-visibility.ts`).

PR #1316, opened 2026-07-24
release: promote dev to production

krishagel:
## Production promotion

Promotes the currently deployed and verified dev release to production.

Includes:
- Atrium upgrades #1284–#1290
- Production OIDC/CDK readiness fixes from #1304 and #1306
- Admin hub update from #1307
- Database migrations 126–131

Dev verification completed:
- CloudFormation UPDATE_COMPLETE
- ECS rollout completed and ALB target healthy
- /api/health healthy, including database and OAuth signing
- OIDC discovery and JWKS return 200 with S256 and RSA/RS256 key
- Migrations 126–130 completed with expected schema objects
- Active content tool catalog v3 rows contain optional sourceRef; prior versions retired
- No post-deploy OIDC, catalog immutability, or severe application errors
- Current dev CI, CDK validation, PostgreSQL lifecycle, CodeQL, and OpenWiki workflows green

Production deployment order after merge: database stack first (migrations 126–131), then frontend ECS stack.

PR #1318, opened 2026-07-24
fix(oauth): promote signing-key rotation cache fix

krishagel:
Production follow-up to #1316.

Promotes the P1 OIDC provider-cache fix from #1317 so scheduled/incident signing-key rotation switches the active signing key without requiring an ECS restart.

Validation:
- Focused OIDC tests: 9 passed
- Typecheck passed
- No database migration or CDK resource change


krishagel:
@claude Please review this narrowly scoped P1 signing-key rotation cache fix and provide a written result.

PR #1319, opened 2026-07-24
fix(tools): version published catalog schema updates

krishagel:
Fixes production startup immutability violations found during the post-deploy audit.

Bumps the published contracts for:
- assistants.execute and assistants.list to v2
- chat.show_chart, chat.web_search, chat.code_interpreter, and chat.generate_image to v2

Older catalog rows remain retired; the current schemas are inserted under new immutable versions. Generated OpenAPI is updated for the REST assistant tools.

Validation:
- Catalog sync/version/OpenAPI tests: 37 passed
- Typecheck passed
- OpenAPI generated-file check passed
- No migration or CDK resource change

PR #1320, opened 2026-07-24
fix(tools): promote catalog immutability cleanup

krishagel:
Production follow-up to the post-deploy audit.

Promotes #1319 so six historically drifted v1 tool schemas are published as immutable v2 contracts. No migration or CDK resource change; redeploying the frontend stack will sync the new versions and retire v1.


krishagel:
@claude Please review this narrow production promotion for the six immutable catalog version bumps and provide a written result.

PR #1335, opened 2026-07-25
feat: #1330 - Nexus conversation auto-deletion (retention setting + Keep flag)

### LINKED ISSUE #1330: Nexus conversation auto-deletion: retention setting + Keep flag
## Summary

Nexus conversations currently accumulate forever — the UI "Delete" button only archives (`archiveConversationAction` → `is_archived = true`), and the unused `deleteConversation()` helper has zero callers. This feature adds admin-configurable retention-based auto-deletion: a nightly scheduled Lambda in **DatabaseStack** hard-deletes conversations whose last message is older than `NEXUS_CONVERSATION_RETENTION_DAYS`, unless the user has flagged the conversation with a new **Keep** toggle (`is_saved` column). Archived-but-stale conversations ARE deletable. The feature ships **disabled** (setting seeded empty/0 = sweep no-ops); Hagel enables it manually in the admin settings UI after user communications go out.

Decided design points (do not re-litigate):
- Keep = new `is_saved` boolean column on `nexus_conversations` + a Keep toggle in the conversation list UI (not metadata, not reuse of `is_pinned`).
- Staleness clock = `last_message_at` only (verified: it is written only on message persistence in `app/api/nexus/chat/chat-helpers.ts` `saveUserMessage`/`saveAssistantMessage`/`saveConversationSteps` and `image-generation-handler.ts`; it is NOT touched on read/open — safe inactivity clock). Opening/reading a conversation does not reset the clock.
- Archived conversations are eligible for deletion (archive is not protection; Keep is).
- Deletion is a **hard delete** (row + cascades + S3 cleanup), not a soft delete.
- Infra home = **DatabaseStack** (`infra/lib/database-stack.ts`), new scheduled Lambda modeled on the `AgentTelemetryPrune` pattern in `infra/lib/agent-platform-stack.ts` (~L3860–3955).
- Conservative default: pinned conversations are also excluded from deletion (`is_pinned = true` treated like Keep). Pinning is an explicit user signal and hard delete is irreversible; flipping this later is a one-line predicate change.

## Definition of Done

- [ ] Migration `135-nexus-conversation-retention.sql` adds `is_saved BOOLEAN NOT NULL DEFAULT false` to `nexus_conversations`, adds a partial index supporting the sweep scan (e.g. on `last_message_at` `WHERE is_saved = false AND is_pinned = false` — exact shape implementer's choice), and seeds a `settings` row `NEXUS_CONVERSATION_RETENTION_DAYS` with a disabled value (`''` or `'0'`) via `INSERT ... ON CONFLICT DO NOTHING`; contains NO `DO $$` blocks; filename appended to `migrationFiles` in `infra/database/migrations.json`; applies cleanly on local Docker DB (`bun run db:reset`)
- [ ] Drizzle schema (`lib/db/schema/tables/nexus-conversations.ts`), `updateConversation()` (`lib/db/drizzle/nexus-conversations.ts`), and `PATCH /api/nexus/conversations/[id]` accept/persist `isSaved`; conversation list payloads include `isSaved`
- [ ] Keep toggle in the Nexus conversation list (`components/nexus/conversation-list.tsx` `ConversationItemRow`) toggles `isSaved` via the existing PATCH route and shows a visible "kept" indicator on saved conversations
- [ ] Eligibility predicate is exactly: `is_saved = false AND is_pinned = false AND last_message_at < now() - retention_days` (archived rows included; kept or pinned rows never deleted regardless of age or archive state) — covered by unit tests on an extracted, pure predicate/SQL-builder module
- [ ] Disabled-by-default is enforced: when `NEXUS_CONVERSATION_RETENTION_DAYS` is missing, empty, `0`, negative, or non-numeric, the sweep exits as a no-op without deleting anything — covered by unit tests
- [ ] Hard delete of an eligible conversation removes: the `nexus_conversations` row (cascading `nexus_messages`, `nexus_conversation_events`, `nexus_conversation_folders`, `nexus_cache_entries`, `nexus_shares`, `nexus_provider_metrics`, `nexus_repository_bindings`); the conversation's ephemeral `knowledge_repositories` + `repository_items` rows and their S3 storage (repository IDs resolved via `nexus_repository_bindings` BEFORE the conversation row is deleted — after cascade they are unreachable); and legacy `documents` rows with `conversation_id` = the conversation (plus their S3 objects at `documents.url`) instead of leaving SET-NULL orphans. Message-part `s3Key` objects are deleted best-effort (log-and-continue). No orphaned `knowledge_repositories` rows remain after a sweep (asserted in tests)
- [ ] New scheduled Lambda (`infra/lambdas/nexus-conversation-retention/`) wired into `DatabaseStack`: role via `ServiceRoleFactory.createLambdaRole` (`vpcEnabled: false` + manually attached `service-role/AWSLambdaVPCAccessExecutionRole`, per the AgentTelemetryPrune pattern), explicit LogGroup named exactly `/aws/lambda/<functionName>`, ARM_64, `reservedConcurrentExecutions: 1`, VPC `PRIVATE_WITH_EGRESS`, DB secret read grant, nightly EventBridge cron rule, per-run batch cap via env var; supports a `{ "dryRun": true }` event that returns candidate conversation IDs/counts without deleting; `bunx cdk synth` passes from `/infra`
- [ ] Full test suite green
- [ ] Zero lint warnings; typecheck clean
- [ ] E2E flow(s) pass — flows: `nexus-keep-toggle`
- [ ] (manual) Post-deploy in dev: invoke the Lambda with `{"dryRun": true}` and confirm candidate list is plausible before enabling; retention value set in admin settings only after user comms (Hagel does this — the PR must NOT enable it)

## Acceptance tests / E2E flows
- `nexus-keep-toggle` (authenticated functional spec, gated by `PLAYWRIGHT_AUTH_ENABLED`, using `tests/e2e/helpers/session-auth.ts` `authenticateContext` against the host `:3100` dev server — see `tests/e2e/nexus/organization.spec.ts` and `tests/e2e/atrium-delete.functional.spec.ts` for the pattern): create a conversation via `POST /api/nexus/conversations`; open Nexus; toggle Keep from the conversation item's menu; assert `GET /api/nexus/conversations` returns `isSaved: true` for it and the kept indicator is visible in the list; toggle off and assert it clears. Screenshot the kept state.
- Sweep logic is a scheduled Lambda and is not Playwright-exercisable; it is covered by the unit tests named in the DoD plus the manual dry-run item.

## Task breakdown
1. **Migration 135** — `infra/database/schema/135-nexus-conversation-retention.sql`: `ALTER TABLE nexus_conversations ADD COLUMN is_saved BOOLEAN NOT NULL DEFAULT false;` + partial index + settings seed row (`NEXUS_CONVERSATION_RETENTION_DAYS`, disabled value, description stating "empty/0 disables auto-deletion; N = days since last message", category matching the existing `NEXUS_ATTACHMENT_RETENTION_DAYS` entry). Append to `infra/database/migrations.json` `migrationFiles`. Verify: `bun run db:reset` applies it locally.
2. **Drizzle + API** — add `isSaved` to `lib/db/schema/tables/nexus-conversations.ts`; extend `updateConversation()` and the `PATCH` body handling in `app/api/nexus/conversations/[id]/route.ts` (alongside existing `isPinned`/`isArchived`); include `isSaved` in `getConversations()` select (`lib/db/drizzle/nexus-conversations.ts`). Verify: typecheck + existing organization spec still green.
3. **Keep toggle UI** — `components/nexus/conversation-list.tsx`: add `isSaved` to `ConversationItem` (~L27–38), add Keep/Unkeep control in `ConversationItemRow` next to the existing pin/archive/delete controls, optimistic PATCH, kept indicator icon. Verify: E2E flow `nexus-keep-toggle`.
4. **Retention sweep Lambda** — `infra/lambdas/nexus-conversation-retention/`: standalone TS handler (model on `infra/lambdas/agent-telemetry-prune/`); reads DB creds from Secrets Manager; reads `NEXUS_CONVERSATION_RETENTION_DAYS` from the `settings` table each run (no cache — admin changes take effect next run); no-op when disabled/invalid; selects eligible conversations oldest-first with per-run cap; per conversation: resolve ephemeral repo IDs from `nexus_repository_bindings` → collect legacy `documents` (id, url) → collect message-part `s3Key`s → delete S3 objects (repo storage prefixes via the same prefix-delete approach as `lib/repositories/content-platform/storage-cleanup.ts` / `deleteRepositoryObjectVersionsByPrefix`) → delete `knowledge_repositories` + `documents` rows → delete the conversation row (cascades the rest). Structured logging of counts; `dryRun` mode. Extract the eligibility predicate/SQL builder as a pure function with unit tests that run in the main CI suite.
5. **DatabaseStack wiring** — `infra/lib/database-stack.ts`: role (ServiceRoleFactory pattern above), LogGroup, Lambda (ARM_64, bundled like AgentTelemetryPrune's `tryBundle`), EventBridge nightly cron rule, env vars (`ENVIRONMENT`, `DATABASE_HOST`, `DATABASE_SECRET_ARN`, `DATABASE_NAME`, batch cap), S3 permissions scoped to the documents/repository bucket prefixes it must delete from. Verify: `bunx cdk synth` from `/infra`.
6. **Admin settings registry** — add `NEXUS_CONVERSATION_RETENTION_DAYS` to `commonSettings` in `app/(protected)/admin/settings/_components/settings-form.tsx` (next to the existing `NEXUS_ATTACHMENT_RETENTION_DAYS` entry, ~L86) so it appears as a known setting with description. Verify: admin settings page renders it.
7. **Tests** — unit tests for predicate + disabled-config no-op + orphan-free delete (local Postgres where the suite supports it); E2E `nexus-keep-toggle` functional spec; run full suite, lint, typecheck.

## Affected areas
- `infra/database/schema/135-nexus-conversation-retention.sql` (new), `infra/database/migrations.json`
- `lib/db/schema/tables/nexus-conversations.ts`, `lib/db/drizzle/nexus-conversations.ts`
- `app/api/nexus/conversations/[id]/route.ts`
- `components/nexus/conversation-list.tsx`
- `infra/lambdas/nexus-conversation-retention/` (new), `infra/lib/database-stack.ts`
- `app/(protected)/admin/settings/_components/settings-form.tsx`
- `tests/e2e/` (new `nexus-keep-toggle` functional spec)

## Out of scope
- Enabling retention in any environment (manual, post-comms — Hagel only)
- User-facing warnings/countdowns/banners about upcoming deletion (comms handled outside the app)
- Changing the existing "Delete" button semantics (it archives today; leave as-is)
- Backfilling `is_saved` for existing conversations (default false is intended)
- Per-user or per-role retention overrides
- Retention for non-Nexus surfaces (Assistant Architect executions, legacy chat)
- A dedicated bespoke admin card for the setting (plain settings row is sufficient)

## Risk & rollback
- Risk: **high** — hard deletion is irreversible and touches S3 + multiple cascading tables. Mitigations: ships disabled (empty/0 seed → guaranteed no-op), `dryRun` event verified in dev before enabling, Keep + pinned exclusions, per-run batch cap, `reservedConcurrentExecutions: 1`, oldest-first ordering, structured per-run logging of deleted counts.
- Rollback: clear/zero `NEXUS_CONVERSATION_RETENTION_DAYS` in admin settings → sweep no-ops immediately on next run (no deploy needed). Full removal = delete the Lambda/rule from DatabaseStack and redeploy. The `is_saved` column and seed row are additive and can remain (migrations are forward-only). Already-deleted conversations are only recoverable from Aurora snapshots (7-day dev / 30-day prod retention).

krishagel:
## Summary
Implements #1330. Adds admin-configurable retention-based auto-deletion for Nexus conversations, plus the user-facing **Keep** flag that exempts a conversation from it.

**Ships disabled.** Migration 137 seeds `NEXUS_CONVERSATION_RETENTION_DAYS` with an empty value, and the sweep exits as a no-op for missing/empty/`0`/negative/non-numeric values. Nothing is deleted until Hagel sets a value in `/admin/settings` after user comms. This PR does **not** enable it in any environment.

## Changes

**Migration 137** (`137-nexus-conversation-retention.sql`) — the issue said 135; that number was taken by `135-atrium-asset-initiation-idempotency.sql`, and 136 was then taken by `136-google-content-connectors.sql` (PR #1333) while this branch was in flight, so this rebased onto dev and renumbered to 137.
- `is_saved BOOLEAN NOT NULL DEFAULT false` on `nexus_conversations`.
- Partial index on `(last_message_at) WHERE is_saved = false AND is_pinned IS NOT TRUE`. `IS NOT TRUE` rather than `= false` because `is_pinned` is **nullable** — `NULL = false` is `NULL`, which would have silently dropped those rows from both the index and the sweep. `EXPLAIN` on the local DB confirms the planner matches this index to the sweep query, with `is_saved`/`is_pinned` absorbed into the index predicate rather than left as a residual filter.
- Seeds the retention setting empty (disabled). No `DO $$` blocks (the migration runner's splitter cannot parse them).

**Keep flag end-to-end** — Drizzle schema, `updateConversation()` (including its `.returning()`), `getConversations()` select, and `PATCH /api/nexus/conversations/[id]`. `isSaved` **and** `isPinned` are validated strictly as booleans (400 otherwise): both gate irreversible deletion, so a truthy string must not silently become "kept". `metadata` is now validated as a plain object too (previously written through unvalidated into a JSONB column).

**Keep toggle UI** — bookmark toggle + visible "Kept" badge in the conversation list. This is the component's first PATCH caller. The optimistic flip is **rolled back** on any failure — the UI must never claim a conversation is protected when the server didn't record it.

**Retention sweep Lambda** (`infra/lambdas/nexus-conversation-retention/`) wired into `DatabaseStack` — nightly 05:00 UTC, ARM_64, `reservedConcurrentExecutions: 1`, VPC `PRIVATE_WITH_EGRESS`, per-run batch cap of 200, `{"dryRun": true}` support.

### Safety properties worth reviewing
- **Claim before destroy.** The guarded conversation-row `DELETE` is the **first** destructive act, and its `WHERE` re-asserts the **full** eligibility predicate — Keep, pin *and* the age cutoff. Until that claim succeeds nothing has been deleted: not the row, not S3 storage, not a repository. A user who clicks Keep, pins, or simply sends a new message at any point before the claim keeps the conversation and every byte of its data. An earlier revision of this PR deleted repository storage before the gate, which honoured the flag on the row but had already destroyed bound repository objects when the user won the race.
- **Two-layer race protection.** `findCandidates` snapshots up to 200 conversations at the top of a run and the loop then works through them sequentially over what can be minutes. A late `isStillEligible` re-read against live state runs before any destructive step, and the guarded `DELETE` closes the remaining window atomically.
- **Resolve-before-delete.** `nexus_repository_bindings` cascades from the conversation, so ephemeral repository IDs are resolved *before* the row is deleted — otherwise the `knowledge_repositories` rows orphan permanently. Covered by an ordering test.
- **Ephemeral repositories only.** A conversation can also be bound to a **promoted** (durable) repository the user deliberately kept. `getBoundRepositoryIds` filters to `repository_kind = 'ephemeral'` at the SQL level, so the sweep cannot destroy exactly the data the promotion flow exists to preserve.
- **Version-aware deletes.** The documents bucket is versioned and `storage-stack.ts` scopes its noncurrent-version lifecycle rules to `repositories/` and `atrium/pending-assets/` only — nothing covers `conversations/` or `v2/generated-images/`. A plain `DeleteObject` would add a delete marker and leave "permanently deleted" content recoverable, so every delete pages `ListObjectVersions` and removes versions *and* delete markers. A truncated listing without a cursor throws rather than silently stopping half-way.
- **Conversation storage is swept by prefix, not by message part.** `chat-helpers.ts` persists user image parts as `{ type: 'image', metadata: { hasImage: true } }`, dropping `s3Key` from the row — those objects have nothing left in the database pointing at them, so a parts-driven delete would orphan them permanently.
- **Legacy `documents.url`.** Rows written before commit `49230981` (Jun 2025, `url: fileUrl` → `url: s3Key`) hold presigned URLs; those are parsed only when the host proves the object is in this deployment's bucket, else the row is deleted without touching S3 and `document_url_unresolvable` is logged.
- **IAM backstop.** The object-level delete grant has to stay bucket-wide (legacy `documents.url` keys are `<userId>/<timestamp>-<file>`, a shape no IAM pattern expresses), so an explicit `Deny` covers namespaces this sweep must never touch: `atrium/`, `public-images/`, `v2/uploads/`, `v2/generated-documents/`.
- **Stack ordering.** `DatabaseStack` now reads StorageStack's `documents-bucket-name` SSM parameter. That is a CloudFormation dynamic reference, not an `Fn::ImportValue`, so CDK cannot infer the dependency — declared explicitly in `infra/bin/infra.ts` for both environments (StorageStack has no reverse reference, so it cannot cycle).

## Verification
Run on the rebased branch at `37b693a6`:
- Typecheck: ✅ app + `infra`
- Lint: ✅ zero warnings in every touched file (the repo carries ~399 pre-existing warnings elsewhere; none are in this diff). The one pre-existing `complexity` warning in a touched file — `POST /api/nexus/messages`, 21 vs max 15 — was refactored below the threshold rather than left.
- Tests: ✅ 353 Jest suites / 3492 tests, plus 55 `bun:test` cases for the sweep
- `cdk synth AIStudio-DatabaseStack-Dev`: ✅ (Lambda bundles; verified ARM_64, reserved concurrency 1, LogGroup named exactly `/aws/lambda/<functionName>`, nightly cron, and the IAM statements)
- Migration: ✅ applies cleanly and **idempotently** on local Docker Postgres (re-run is a no-op)
- E2E flow `nexus-keep-toggle`: ✅ deterministic — verified `--repeat-each=3 --retries=0`, 6/6

**Full E2E suite — honest status.** The full pre-push suite does **not** pass cleanly on this machine, and *neither does `dev`*. Measured back to back:
- this branch: 3 failed, 3 flaky, 262 passed
- clean `origin/dev` (`2c5d207c`): 1 failed, 1 flaky, 264 passed — a **different** spec

Every spec that failed on this branch (`unified-content-product-migration.functional`, `atrium-meridian-editor`, `assistant-architect-model-routing`, `agent-workspace-connect`, `admin-users`) passes in isolation on this branch. The failure set differs run to run and across branches, so this is load/ordering flakiness on the local runner, not a regression from this PR. The push therefore used the documented `SKIP_E2E=1` bypass. Flagging rather than burying it — CI is the authority here.

The new Lambda's tests were added to `.psd/verify.json`: `jest.config.js` ignores `/infra/` outright, so they would otherwise never have gated anything.

## Evidence
Kept state — filled bookmark toggle + "Kept" badge on the kept row, unkept rows alongside:

nexus-keep-toggle

## Notes for review — accepted trade-offs, flagged not hidden
- **Repository-storage failure keeps the row.** After the claim, if a repository prefix delete fails, that repository's `knowledge_repositories` row is deliberately **kept** so row and objects stay consistent as a pair (ERROR-logged). This is a deliberate reading of the issue's "no orphaned `knowledge_repositories` rows" criterion: dropping the row anyway would strand real bulk data in S3 with nothing pointing at it, whereas a kept row is still reclaimable through the ephemeral-repository lifecycle (`expires_at` / `lifecycle_status`). Worth an explicit ack.
- **Best-effort object deletes.** Conversation-prefix, legacy-document and message-part deletes are best-effort after the claim, so a transient S3 error can leave orphaned objects once the row is gone. Blocking on them would make the sweep trivially stallable.
- **Non-concurrent index build.** `CREATE INDEX CONCURRENTLY` cannot run through the RDS Data API migration runner, so this matches every other migration here and takes a `SHARE` lock for the build. `ADD COLUMN` with a constant default is catalog-only on PG11+. Check prod `nexus_conversations` size before deploying.
- **`stripClientStorageRefs` is now a deletion boundary**, not only a presign one — the sweep permanently destroys the objects named by `parts[].s3Key`. Documented in the route so a future `nexus_messages` writer cannot quietly reopen it.

## Post-deploy (manual, Hagel)
1. Invoke the Lambda with `{"dryRun": true}` and confirm the candidate list is plausible.
2. Only then set `NEXUS_CONVERSATION_RETENTION_DAYS`, after user comms.

Closes #1330


krishagel:
Thanks — all three P1s were real. They were reviewed against `31aa4448e3`, which predates the rebase onto dev; every one of them is fixed at `37b693a6`. Verification for each:

**1. Exclude promoted repositories from the retention sweep** — fixed. Bound-repository resolution now filters at the SQL level and is pinned as an exported constant so it cannot silently regress (`index.ts`, `BOUND_EPHEMERAL_REPOSITORIES_SQL`):

```sql
SELECT b.repository_id
FROM nexus_repository_bindings b
JOIN knowledge_repositories r ON r.id = b.repository_id
WHERE b.conversation_id = $1::uuid
  AND r.repository_kind = 'ephemeral'
```

`durable` (promoted via `promoteNexusRepository`) and `system` repositories keep their conversation binding and now survive the sweep, which was exactly your point.

**2. Recheck message age in the atomic DELETE** — fixed. The DELETE re-asserts the *full* eligibility predicate, not just the flags, reusing the same `CANDIDATE_WHERE_CLAUSE` constant as the batch scan so the two cannot drift:

```sql
DELETE FROM nexus_conversations
WHERE <CANDIDATE_WHERE_CLAUSE>   -- is_saved / is_pinned / last_message_at < cutoff
  AND id = $2::uuid
RETURNING id
```

A user who sends a new message mid-sweep bumps `last_message_at`, the WHERE matches nothing, 0 rows come back, and the caller aborts.

**3. Gate repository deletion on the Keep decision** — fixed, and this one changed the ordering rather than just the predicate. The guarded row DELETE is now the **first** destructive act; all storage cleanup (conversation prefixes, legacy objects, repository prefixes) runs only *after* the claim succeeds. Previously the claim honoured the flag on the row but repository objects were already gone — the Keep guarantee still ended in irreversible data loss, precisely as you described. Now losing the race costs the user nothing at all.

Also worth flagging for the next pass: repository-storage failure after the claim deliberately **keeps** that repository's `knowledge_repositories` row so row and objects stay consistent as a pair (the ephemeral lifecycle is the reclaim path), rather than dropping the row and stranding bulk data in S3.

Coverage: 55 `bun:test` cases now, including the ordering assertion that the re-check precedes any storage deletion, and that a conversation which loses the race has nothing deleted — not the row, not S3, not a repository.

@codex review

krishagel:
Both new P1s were real and are fixed at `cfed09b9`.

**1. Revalidate repository kind before destructive cleanup** — correct, resolution-time filtering could not cover that interleaving. Fixed by applying the same claim-before-destroy shape already used for the conversation itself: `deleteRepositoryRows` became `claimRepositoryRows`, whose DELETE re-asserts the kind and returns what it actually claimed:

```sql
DELETE FROM knowledge_repositories
WHERE id = ANY($1::int[])
  AND repository_kind = 'ephemeral'
RETURNING id
```

Storage is then swept **only** for the returned IDs. A repository promoted to `durable` since resolution matches nothing, is not returned, and keeps both its row and every byte of its storage (logged `repository_no_longer_ephemeral`).

This inverts the previous row/storage ordering, which is a deliberate trade: a failed prefix delete now orphans objects (ERROR-logged) rather than stranding a row. Orphaned objects are recoverable operationally; a promoted repository destroyed mid-sweep is not.

**2. Restrict legacy keys to conversation-owned storage** — also correct, and your point that sanitising the current writer says nothing about historical rows is the crux. Out-of-prefix `parts[].s3Key` values are now accepted only inside the **owner's** legacy upload namespace `<userId>/` (the shape `uploadDocument()` builds when `repositoryId` is null); anything else is skipped. An orphaned legacy object is the pre-existing status quo and is strictly better than deleting an object that belongs to someone else. The owner id is plumbed through the port from the candidate row.

`documents.url` is deliberately left as-is: ownership there is established by the row itself (`conversation_id` + `user_id` FKs) and the value is server-written by the upload route, unlike a free-form key inside a jsonb blob.

Coverage is now 61 `bun:test` cases, adding:
- a repository promoted mid-sweep keeps BOTH its row and its storage
- five ownership cases for `extractOutOfPrefixKeys`, including that user `7` cannot reach user `70`'s namespace via a string-prefix match

Verified on the pushed commit: 353 Jest suites / 3492 tests, app + `infra` typecheck, `cdk synth`, and the `nexus-keep-toggle` E2E flow.

@codex review

krishagel:
Confirmed and fixed at `078f0c6f` — thanks, this one was subtle and real.

`upsertMessageWithStats` (`lib/db/drizzle/nexus-messages.ts:544-552`) does exactly what you described, and says so itself:

> Note: Not atomic - stats may be briefly out of sync if update fails.

So a resumed conversation can have a committed `nexus_messages` row while `last_message_at` still reads stale — transiently between the two statements, and *permanently* if the stats update failed. Gating an irreversible delete on that denormalized column alone would cascade away the freshly committed message.

Both authoritative gates — the late `isStillEligible` re-check and the claiming DELETE — now also test the source of truth:

```sql
NOT EXISTS (
  SELECT 1 FROM nexus_messages m
  WHERE m.conversation_id = nexus_conversations.id
    AND m.created_at >= (now() AT TIME ZONE 'UTC') - ($1::int * interval '1 day')
)
```

Kept out of the bulk candidate scan on purpose: that stays a cheap indexed range read over `last_message_at`, and the per-row check runs only where it actually decides something. `EXPLAIN` on the guarded DELETE confirms it costs one index lookup per claimed conversation rather than a scan:

```
Delete on nexus_conversations
  -> Nested Loop Anti Join
       -> Index Scan using idx_nexus_conversations_retention_sweep
       -> Index Scan using idx_nexus_messages_conversation
            Index Cond: (conversation_id = $2 AND created_at >= ...)
```

`idx_nexus_messages_conversation (conversation_id, created_at)` already existed from migration 031, so no new index is needed.

Coverage is now 64 `bun:test` cases, pinning the fragment's shape, its UTC comparison and `$1` reuse, and asserting the bulk scan does **not** carry it.

Verified on the pushed commit: 353 Jest suites / 3492 tests, app + `infra` typecheck, and the `nexus-keep-toggle` E2E flow.

@codex review

krishagel:
All three fixed at `5c29c4a5`.

**P1 — Claim linked documents before deleting their storage.** Confirmed: `linkDocumentToConversation` relinks by id alone with no old-conversation check, so a document resolved earlier in the sweep can since belong to a different, live conversation.

`deleteConversationRow` + `deleteDocumentRows` are now a single `claimConversation` port implemented as ONE transaction, with a forced order:

1. `DELETE FROM documents WHERE conversation_id = $conv RETURNING id, url` — **must** be first, because the conversation DELETE cascades that column to NULL and would erase the very link the guard depends on.
2. The guarded conversation DELETE (full predicate).
3. If (2) matches nothing, a `ClaimLostError` sentinel is thrown so postgres.js rolls the document deletes back with it — losing the race still costs the user nothing.

Storage deletion is driven off the documents the claim actually `RETURNING`ed, so a relinked document keeps both its row and its object (logged `document_relinked_mid_sweep`). `EXPLAIN` confirms the guarded delete uses `idx_documents_conversation_id`.

**P2 — Reject retention values outside the executable range.** Correct, and the failure mode is worse than "not executable": `2147483648` overflows the `$1::int` bind, and a large enough window makes `retentionCutoff()` produce an Invalid Date whose `.toISOString()` throws — so the nightly sweep becomes a hard error rather than the no-op an operator would expect from a bad setting. Now bounded at `MAX_RETENTION_DAYS` (36500 — 100 years) with a new `too_large` reason, failing closed like every other unusable value.

**P2 — Track Keep requests per conversation.** Also correct. `togglingKeepId` is now a `Set` keyed by conversation id. With the single id, starting a request for B re-enabled A's button, and whichever settled first cleared the id while the other was still in flight — allowing a second toggle whose PATCH and optimistic rollback could resolve out of order and leave the UI disagreeing with the persisted protection state.

Coverage is now 67 `bun:test` cases, adding a relinked document keeping row *and* storage, and the retention upper bound (including that the boundary itself still yields a valid cutoff).

Verified on the pushed commit: 353 Jest suites / 3492 tests, app + `infra` typecheck, zero lint warnings in touched files, and the `nexus-keep-toggle` E2E flow.

@codex review

krishagel:
Both fixed at `d3a2d07b`.

**P1 — Lock the conversation before checking for new messages.** Right, and the committed-message check alone could not close this — it needed a lock. The claim transaction now takes this as its **first** statement:

```sql
SELECT id FROM nexus_conversations WHERE id = $1 FOR UPDATE
```

`FOR UPDATE` conflicts with the `FOR KEY SHARE` a message INSERT takes on the referenced conversation row, so the sweep waits for any in-flight insert to finish. And because each statement under READ COMMITTED takes a fresh snapshot, the subsequent DELETE then *sees* that committed message and matches nothing. Previously the DELETE could evaluate `NOT EXISTS` on a pre-insert snapshot, block on the FK lock, and delete after the insert committed — cascading away a message it never saw.

**P2 — Filter recent committed messages before limiting candidates.** Also right, and the consequence is worse than inefficiency: with `upsertMessageWithStats` explicitly non-atomic, a permanently-failed stats update leaves a stale `last_message_at` forever, so those rows are re-selected every night, skipped by the gate, and hold batch slots. 200 of them ahead of the real backlog and `ORDER BY last_message_at ASC LIMIT` starves the sweep entirely. The candidate scan now composes the same fragment, so it only returns genuinely eligible rows.

Cost is bounded — `EXPLAIN` confirms it stays an index-driven anti-join that stops at the LIMIT, not a sequential scan:

```
Limit
  -> Nested Loop Anti Join
       -> Index Scan using idx_nexus_conversations_retention_sweep
       -> Bitmap Index Scan on idx_nexus_messages_conversation
```

The fragment stays a separate constant composed at each call site rather than merged into `CANDIDATE_WHERE_CLAUSE`, so the migration-137 partial-index predicate still matches the scan's own WHERE exactly (there is a test pinning that).

Verified on the pushed commit: 67 `bun:test` cases, 353 Jest suites / 3492 tests, app + `infra` typecheck, `cdk synth`, and the `nexus-keep-toggle` E2E flow.

@codex review

krishagel:
Fixed at `00189c49` — and it was worse than the Keep button alone.

The row is a `role="button"` with its own `onKeyDown` calling `onSelect` for Enter/Space. Those keys bubble from any nested control, so this already affected the **existing** Delete button too: Enter on it opened the confirm dialog on top of a navigation. The Keep toggle just made it two-for-two.

Guarded by origin rather than by key:

```ts
if (e.target !== e.currentTarget) return
```

so the row only responds to keys aimed at the row itself. Mouse behaviour is unchanged — the buttons already `stopPropagation` on click.

Added an E2E case that focuses the Keep button, presses Enter, asserts the PATCH returns 200 and `aria-pressed` flips, and asserts `page.url()` is unchanged — the URL assertion is what actually pins the regression. The `nexus-keep-toggle` flow is now 3/3.

Verified on the pushed commit: 353 Jest suites / 3492 tests, typecheck and lint clean.

@codex review

Issue #1305, opened 2026-07-24
Agent user-slot Drive: read + organize (drive.readonly + drive.metadata), folder-create only, delete impossible


krishagel:
## Summary
Users want the agent to **read** and **organize** their own Google Drive. Design approved by Kris (2026-07-25): extend the **user slot** (`user_account` OAuth consent) with `drive.readonly` + `drive.metadata`, and allow **folder creation only** through the already-granted `drive.file`. File creation in the user's Drive stays banned (creation continues to happen on the `agnt_` account + explicit share), and delete is made **scope-impossible**, not merely gate-blocked.

This consciously refines the 2026-07-06 product decision (the `drive.file`-only barrier): the *impersonated-creation* ban stays; read + organize are now in scope, folders explicitly allowed, files explicitly not.

## Target scope set (user slot)
`gmail.modify`, `calendar`, `tasks`, `drive.file` (existing) **+ `https://www.googleapis.com/auth/drive.readonly` + `https://www.googleapis.com/auth/drive.metadata`** (new)

## Enforcement model (layered)
| Operation | Allowed? | Enforced by |
|---|---|---|
| Read/list/export any user file | ✅ | `drive.readonly` |
| Rename, move (add/removeParents), star | ✅ | `drive.metadata` |
| Create **folder** in user Drive | ✅ | `drive.file` (`files.create` `mimeType=application/vnd.google-apps.folder`) + gate carve-out allowing ONLY the folder mimeType |
| Create **file/doc/content** in user Drive | ❌ | Skill gate (existing USER_SCOPE_FORBIDDEN keeps blocking non-folder `files.create`/`copy`) — impersonation barrier per 2026-07-06 |
| Modify file **content** | ❌ | **Google** — neither new scope permits content writes/uploads |
| Trash (`trashed=true`) | ❌ | Skill gate (existing trash blocks). VERIFY whether `drive.metadata` even permits it (trashed is a metadata field); document which layer holds |
| Permanent delete (`files.delete`, `emptyTrash`) | ❌ | **Google** — requires `drive`/`drive.file` on the target; impossible for non-agent-created items under this scope set |

Rejected alternatives (for the record): full `drive` + regex-only gating (delete depends on our gate — unacceptable for "absolutely no delete"); DWD impersonation of real users (consent-less, breaks the `agnt_`-only containment invariant of #1232); share-folders-to-`agnt_` status quo (organizing needs Editor = content-edit power, coverage gaps).

[...]

PR #1351, opened 2026-07-26
feat: #1239 - resolve district identities from email and Chat user id

### LINKED ISSUE #1239: Agent directory identity resolution: resolve district emails + Chat IDs to real people (enable People API)
## Goal

The agent should resolve identities instead of guessing:
- **Email address → who they really are in the district** (name, title, org unit).
- **Google Chat `users/{id}` → who they really are** (map the Chat sender/mention ID to a real person).

Follow-on to the agent workspace broker (#1232/#1233, shipped in #1235). Supersedes the earlier "connection audit" item that landed on `contacts/contacts.readonly` — see correction below.

## Verified findings (Google docs)

**`contacts.readonly` is the wrong scope.** It grants the user's *personal contacts*, not the district directory, so it does not answer "who is this district email / Chat ID." Do **not** add it for this.

**Email → district person:**
- Scope `https://www.googleapis.com/auth/directory.readonly` — **already in `AGENT_DWD_SCOPES`** (`lib/agent-workspace/dwd-token-broker.ts`).
- Mechanism: People API `people.searchDirectoryPeople` (query = email) returns the domain profile.
- **Only missing piece: enable `people.googleapis.com` on the agent GCP project (project number is in the `psd-agent/{env}/gcp-dwd-config` Secrets Manager secret — ask Kris).** (This is the one real item the connection audit surfaced.)

**Chat ID → district person:**
- The Chat `users/{id}` numeric ID is the **same identifier** as the People API Person ID and the Admin SDK Directory user ID — a direct lookup.
- Two paths:
  - **Low-privilege (try first):** People API `people.get` on the directory-profile ID with `directory.readonly` (already granted). Known gotcha: in a service-account context it can return **empty fields**.
  - **Guaranteed fallback:** Admin SDK Directory `users.get`, scope `https://www.googleapis.com/auth/admin.directory.user.readonly`. Google's least-privilege pattern is a custom **"Users > Read"** admin role on the **service account**, calling Admin SDK *as the SA* — NOT via `agnt_` impersonation (an `agnt_` DWD token with this scope would 403, since `agnt_` isn't an admin). So this is a **separate call path**, not an `AGENT_DWD_SCOPES` edit.

## Proposed approach

**Option A (recommended, low-privilege first):**
1. Reese enables `people.googleapis.com` on the agent GCP project (number in the `gcp-dwd-config` secret).
2. Agent/skill uses the existing `directory.readonly` scope to call `searchDirectoryPeople` (email) and `people.get` (Chat ID).
3. Only if the Chat-ID `people.get` returns empty, escalate to Option B for Chat IDs.
- **No new DWD scope, no admin role.** `AGENT_DWD_SCOPES` unchanged.

**Option B (guaranteed Chat-ID path, escalate only if needed):**
- Grant the DWD service account (email in the `gcp-dwd-config` secret) a custom "Users > Read" admin role; enable Admin SDK API; call Admin SDK Directory `users.get` with `admin.directory.user.readonly` (SA-as-itself, separate from the DWD broker).

## Where the code changes land

- Email resolution: **no scope change** — `directory.readonly` is already in `AGENT_DWD_SCOPES`. The work is (a) People API enabled, (b) an agent skill / helper that calls `searchDirectoryPeople` + `people.get` and caches lookups.
- Chat-ID Admin SDK fallback (only under Option B): a new SA-as-itself Admin SDK call path (not `AGENT_DWD_SCOPES`).

## Google-side actions (Reese)

- [ ] Enable `people.googleapis.com` on the agent GCP project (number in the `gcp-dwd-config` secret).
- [ ] (Only under Option B) custom "Users > Read" admin role on the SA + enable Admin SDK API.

## Acceptance criteria

- [ ] Given a district email, the agent returns the person's real name/title from the directory (no guessing).
- [ ] Given a Chat `users/{id}`, the agent returns the real person (name + primary email).
- [ ] `contacts.readonly` is NOT added.
- [ ] Least-privilege path used unless Chat-ID `people.get` empirically returns empty, then Option B.
- [ ] Lookups cached to avoid per-message directory calls.

## Sources

- People API — searchDirectoryPeople (directory.readonly): https://developers.google.com/people/api/rest/v1/people/searchDirectoryPeople
- OAuth 2.0 Scopes for Google APIs: https://developers.google.com/identity/protocols/oauth2/scopes
- Identify and reference Google Chat users: https://developers.google.com/workspace/chat/identify-reference-users
- Resolving Google Chat User IDs to Emails (least-privilege / Admin SDK custom role): https://justin.poehnelt.com/posts/resolving-google-chat-user-ids-to-emails/

---
_Analysis captured via Claude Code: https://claude.ai/code/session_017FwibWMuzAtnMuvHFAYiPo_

krishagel:
Closes #1239

Adds the `psd-directory` skill so the agent asks the directory who someone is instead of guessing from an address local-part or leaving a Chat sender as an opaque `users/{id}`.

## The issue was wrong about the blocker, twice

The **"Google-side actions (Reese)"** section was stale — the broker moved to the district-owned `psd-aistudio-broker` project on 2026-07-16 and `people.googleapis.com` was already enabled there.

The actual blocker was an **admin console setting**: External Directory Sharing was on its restrictive default, so every directory call 403'd regardless of scope. My own first writeup then misnamed it "Contact sharing" (a different setting). It was corrected and toggled on 2026-07-26, and this code was written **only after** a live probe confirmed the surface returns data.

## Option B is deliberately not built

The genuinely open question was whether `people.get` returns populated fields for a Chat id in a **service-account** context, or comes back empty — empty would have forced Option B (Admin SDK `users.get` behind a custom "Users > Read" admin role on the SA).

Probed live with an `agnt_` DWD token: **it returns fully populated fields.** That privilege escalation is unnecessary and is not implemented. `AGENT_DWD_SCOPES` is untouched; `contacts.readonly` is still absent.

## What it does

```bash
run.js --user <owner> --email someone@psd401.net    # searchDirectoryPeople
run.js --user <owner> --chat-id users/11626491...   # people.get, same id
```

Returns one JSON line with `displayName`, `email`, `title`, `department`, `organization`. `organizations` supplies title/department directly, so the "name, title, org unit" criterion is one call, not two.

## Correctness decisions that aren't cosmetic

- **Exact-address match only.** `searchDirectoryPeople` is a prefix/substring search, so one address can return several people. Taking `[0]` would make the agent confidently attribute a message to the **wrong human** — worse than not resolving. A near miss reports `found:false`.
- **A miss is exit 0.** "Not in the directory" is an answer the agent must act on, not an error.
- **An empty-field `people.get` reports `found:false`**, not a person with null everything — that was the exact Option-B trigger, so if it ever starts happening it must be visible.
- **Exit 16 is reserved for the admin-console setting**, exit 17 for a genuinely missing scope. Collapsing those into a generic 403 is what made this issue expensive to diagnose the first time.

## ⚠️ No enumeration, by design — please read

The probe surfaced that the district directory includes **ClassLink-provisioned student records** (name, `@edtools` address, grade level, building), and External Directory Sharing is **domain-wide**. A list/enumerate helper in the agent image would therefore be a student-directory dumper one prompt away.

`listDirectoryPeople` is never called, there is no listing entry point, and a test asserts both against the **comment-stripped** source so the property can't rot silently. SKILL.md explicitly tells the model not to seek a workaround.

Separately: whether student OUs should be scoped out of the shared directory is a live question for Hagel — [custom directories](https://support.google.com/a/answer/9401094) are OU-assignable even though External Directory Sharing itself is not. That's an admin decision, not a code one, and it's noted on the issue.

## Caching (acceptance criterion)

On-disk, container-ephemeral. 12h for hits, 5m for misses — a miss is usually a race with account provisioning, and caching that for hours would make a new hire look permanently unresolvable. Keys are sha256-hashed so no address lands in a filename in plaintext and no key can traverse out of the cache dir. Cache failures never fail a lookup; by then the call already succeeded.

## Testing

- **21 unit tests** (bun), wired into CI as `test:skill:directory`. This skill is plain CJS outside the jest roots, so `test:ci` would never have seen it — the same coverage gap #1305 just fixed for `psd-workspace`.
- **Functionally tested against the live People API** with a real minted token, not only mocks:
  - email → `Kris Hagel / Chief / Multiple Locations`
  - second call served from cache
  - Chat-id path resolved the same person
  - nonexistent address and bogus id both returned clean misses rather than throwing
- Full suite: 364 passed, typecheck clean.

## Deploy note

No Dockerfile change needed — `COPY skills /opt/psd-skills` picks it up, and only skills with dependencies need an npm install entry (this has none). **This skill is not in the image tag pushed earlier today** (`2026-07-26-drive-organize`, which carries the #1305 gate fix); it needs a fresh build-and-push after merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,16 @@ jobs:
       env:
         CI: true
 
+    # psd-directory (#1239). Same story as psd-workspace above: plain CJS
+    # outside the jest roots, so `test:ci` cannot see it. What these tests
+    # guard is identity CORRECTNESS — the exact-address rule that stops a
+    # fuzzy search result being reported as the person, and the absence of any
+    # directory-enumeration entry point.
+    - name: Run psd-directory lookup tests (Bun)
+      run: bun run test:skill:directory
+      env:
+        CI: true
+
     # Optional: Upload test coverage if you generate it
     # - name: Upload coverage reports
     #   uses: codecov/codecov-action@v3
--- a/infra/agent-image/skills/psd-directory/SKILL.md
+++ b/infra/agent-image/skills/psd-directory/SKILL.md
@@ -1,6 +1,8 @@
 ---
 name: psd-directory
+summary: Who is this person? Look up an unknown email address or a Google Chat sender (users/{id}) and get the real person — name, job title, department, school. Identity lookup for staff and colleagues by email, Chat id, or message sender.
 description: Resolve a district email address or a Google Chat `users/{id}` to the real person in the Peninsula SD directory — name, title, department. Use this whenever you are about to refer to someone by a raw email address or a Chat sender id, instead of guessing who they are.
+allowed-tools: Bash(node:*)
 ---
 
 # psd-directory — who is this person, actually
--- a/infra/agent-image/skills/psd-directory/SKILL.md
+++ b/infra/agent-image/skills/psd-directory/SKILL.md
@@ -1,0 +1,92 @@
+---
+name: psd-directory
+description: Resolve a district email address or a Google Chat `users/{id}` to the real person in the Peninsula SD directory — name, title, department. Use this whenever you are about to refer to someone by a raw email address or a Chat sender id, instead of guessing who they are.
+---
+
+# psd-directory — who is this person, actually
+
+You frequently encounter an identity you cannot read: a raw address in a
+message, or a Chat sender that arrives as `users/116264913639920976203`.
+Guessing is the failure mode this skill exists to remove. Ask the directory.
+
+```bash
+# email -> person
+node /opt/psd-skills/psd-directory/run.js --user <owner@psd401.net> --email someone@psd401.net
+
+# Chat users/{id} -> person
+node /opt/psd-skills/psd-directory/run.js --user <owner@psd401.net> --chat-id users/116264913639920976203
+```
+
+`--user` is the person whose `agnt_` account brokers the lookup — normally the
+owner of the conversation you are working in.
+
+## What comes back
+
+One JSON line on stdout.
+
+```json
+{"found":true,"personId":"116264913639920976203","displayName":"Kris Hagel",
+ "email":"hagelk@psd401.net","title":"Chief","department":"Multiple Locations",
+ "organization":"Peninsula School District","cached":false}
+```
+
+A person who is not in the directory is an **answer, not an error** — exit 0:
+
+```json
+{"found":false,"query":"someone@psd401.net","reason":"not in directory"}
+```
+
+When `found` is false, say so. Do not fall back to inferring a name from the
+address local-part; "hagelk@" looking like a Hagel is exactly the guess this
+skill replaces.
+
+## Rules that matter
+
+**Never present a near match as the person.** An email lookup only reports a
+person whose address matches *exactly*. The underlying Google search is a
+prefix/substring search, so a partial address can return several people —
+resolving to the wrong human is worse than resolving to nobody.
+
+**`title` and `department` come from ClassLink OneSync**, so they are as
+current as the last sync, not live HR data. Treat them as descriptive, not
+authoritative, for anything consequential.
+
+**There is no way to list the directory, by design.** This skill resolves
+identities you have already encountered. It cannot enumerate people, and that
+is deliberate: the district directory includes student records (name, address,
+grade, building), so a bulk-listing capability in the agent image would be a
+student-directory dumper one prompt away. If you find yourself wanting "all
+users", stop — that is not a request this tool serves, and it is not one to
+work around by other means.
+
+## Cost and caching
+
+Lookups are cached on disk in the container: 12 hours for a hit, 5 minutes for
+a miss (a miss is often a race with account provisioning, and caching that for
+hours would make a new hire look permanently unresolvable). Repeated lookups of
+the same person inside one session cost nothing — resolve freely rather than
+rationing calls. `--no-cache` forces a fresh read if you have reason to think
+someone was just created or renamed.
+
+## When it fails
+
+| Exit | Meaning | What to do |
+|---|---|---|
+| 0 | Success, including `found:false` | Use the answer |
+| 1 | Bad usage or malformed input | Fix the arguments |
+| 2 | Unexpected People API error | Report it; do not retry blindly |
+| 12 | Broker or People API unreachable | Transient — retry once |
+| 14 | The `agnt_` account is still being created | Tell the user to retry shortly |
+| 16 | **Directory sharing is disabled in the Workspace admin console** | No code or retry fixes this. An admin must set Directory → Directory settings → Sharing settings → External Directory Sharing to "Organization data and authenticated user basic profile fields". Say that plainly rather than reporting a generic permission error. |
+| 17 | The agent token lacks `directory.readonly` | A scope/deploy problem — report it |
+
+Exit 16 is its own code on purpose: it was the real state of this district
+until 2026-07-26, and it looks identical to a permissions bug unless you know
+to name the setting.
+
+## Scope and privilege
+
+Uses `directory.readonly`, which the agent slot already holds — no new OAuth
+scope, no consent flow, no admin role on the service account, and
+`contacts.readonly` is deliberately **not** used (it grants personal contacts,
+which is a different and wrong thing). See issue #1239.
--- a/infra/agent-image/skills/psd-directory/SKILL.md
+++ b/infra/agent-image/skills/psd-directory/SKILL.md
@@ -88,8 +88,8 @@ or renamed, since that one does spend a mint.
 |---|---|---|
 | 0 | Success, including `found:false` | Use the answer |
 | 1 | Bad usage or malformed input | Fix the arguments |
-| 2 | Unexpected People API error | Report it; do not retry blindly |
-| 12 | Broker or People API unreachable | Transient — retry once |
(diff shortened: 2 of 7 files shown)


krishagel:
@codex review — round 2. The P2 is fixed in d753165f, and you were right about the mechanism: `resolveEmail` shaped every candidate *before* the exact-match test, and `shapePerson` reduces `emailAddresses` to the primary — so an alias query returned `found:false` for a person Google had already identified correctly, and cached that miss.

Fix is match-first, shape-second: new `addressesOf()` collects every address on a record and the candidate search tests the normalized query against that full set; only the matched record is shaped. The primary is still reported as the canonical `email`, and a `matchedAlias` field is added when the query hit a non-primary address so the answer cannot look like it is about a different person.

This does not weaken the exact-match rule — it is still equality against a real address on the record, just all of them rather than one — and there is now a test pinning that an alias on the WRONG record still misses.

Worth noting I could **not** reproduce this against live dev: my own directory record carries a single address, so there was nothing to alias-match. I treated it as a structural defect in the code path rather than something dependent on current PSD data, since a record gaining an alias later would have silently broken lookups.

24 tests pass (was 21). The alias test was verified to FAIL against the previous primary-only logic (23 pass / 1 fail) and pass with the fix.

krishagel:
@codex review — round 3. The discoverability P2 is fixed in f0791650, and it was a good catch — I verified both halves of your claim in code rather than taking them on faith: `agent-platform-stack.ts` does substitute `Bundled skill: ${name}` when summary is absent, and `psd-skills-meta/common.js` searches `name ILIKE :q OR summary ILIKE :q` with description never consulted. The skill would have been findable only by someone who already knew the word "directory".

Checking the convention surfaced a second omission your comment didn't mention: the frontmatter also lacked `allowed-tools: Bash(node:*)`. All 30 skills in the image declare it except psd-rules (a rules document, not executable). Without it this skill could not shell out to node — it would not have run at all.

Both added. Discoverability verified by simulating the actual ILIKE against name+summary: email, identity, person, chat sender, who is, directory, title, department, sender, colleague all hit. "chat sender" initially missed because ILIKE needs the phrase contiguous and my draft read "Chat users/{id} sender" — the simulation caught what re-reading the sentence didn't.

New tests parse the frontmatter with the SAME regex the stack uses and assert all four fields plus the search keywords (26 pass, was 24). This defect class is invisible to every other gate: it breaks discovery at runtime, not the build.

Also filed separately: psd-schedules has the identical missing-summary gap today, so scheduling is only findable by the word "schedules". Left out of this PR as out of scope.

krishagel:
@codex review — round 4. The P2 is fixed in e853e889, and I verified the limit you cited: `token-rate-limit.ts` is `RATE_LIMIT_PER_HOUR = 120` over a 1h window, shared with psd-workspace.

You identified the mechanism exactly, and there was a compounding factor worth naming: SKILL.md was telling the model *"repeated lookups of the same person cost nothing — resolve freely rather than rationing calls."* So the documentation was actively steering the agent toward the behavior that exhausts a budget shared with Gmail/Calendar/Drive. A wrong doc plus an eager mint is worse than either alone.

Fix: `resolveEmail`/`resolvePersonId` now accept a token **or** an async provider and resolve it only after the cache misses; run.js passes a memoized `mintToken` closure so one invocation can never mint twice. The not-provisioned and transport paths became typed errors (`ACCOUNT_NOT_PROVISIONED`, `MINT_TRANSPORT`) so exits 14 and 12 are unchanged. SKILL.md now says what is true — a hit costs no call and no mint — and notes that `--no-cache` does spend one.

29 tests (was 26). The mint-avoidance test was verified to FAIL against eager minting (28 pass / 1 fail) and pass with the fix. Also live-verified against the real People API with a counting provider: 3 lookups, 1 mint.

krishagel:
@codex review

(Re-requesting: the previous three pushes each drew a review within 4–8 minutes, but `e853e889` — the lazy-mint fix for your round-3 P2 — has gone ~35 minutes with neither a review nor a 👍, so I don't want to read the silence as a clean verdict.)

Head is `e853e889`, all 7 checks green. Changes since your last review of `f0791650`: broker-token minting moved behind the cache so a hit costs no mint, with the not-provisioned/transport paths converted to typed errors to preserve exits 14 and 12.

krishagel:
@codex review — round 5. The 5xx P2 is fixed in b2bb502b, and following your AGENTS.md:360 citation turned up a second, worse bug in the same function.

**Your finding.** `classifyError` fell through to LOOKUP_FAILED for any non-401/403 status, so an upstream 500/502/503/504 mapped to exit 2 ("do not retry blindly") when SKILL.md promises exit 12 ("transient — retry once") for exactly that case. Callers were told not to retry during the one failure mode retrying is for. 5xx now classifies as TRANSPORT, and the status is the signal rather than the parsed message — a load-balancer 502 carries HTML, so there is nothing in the body to classify on. Covered by a test using that exact shape.

**What the citation surfaced.** AGENTS.md:360 warns against `response.json().catch(() => fallback)` before checking `response.ok`. This function did that, and the consequence on the SUCCESS path was worse than on the error path: a 2xx whose body failed to parse degraded to `{}`, which shaped into `found:false`. A malformed success reported "not in the directory" for a person who is in it — and cached that miss for five minutes. Same silent-wrong-answer class as the alias bug from round 2. Status is now checked first, and an unparseable 2xx throws instead of becoming a false miss.

37 tests (was 34). The two 5xx tests were verified to fail without the mapping (35 pass / 2 fail). SKILL.md's exit table updated to match. Live re-verified: all five paths correct against the real People API.

Note the head is now b2bb502b — it also includes fd5d4b5d, which moved the token minter into lib.js so its exit-14 and exit-12 paths are unit-testable (they previously had no coverage, since run.js can only load inside the image).

PR #1358, opened 2026-07-26
fix: #1239 - move directory lookup server-side behind the agent broker

krishagel:
Follow-up to #1351. Depends on [PR #1356](https://github.com/psd401/aistudio/pull/1356) to build the image.

## psd-directory was broken within a minute of merging

[PR #1353](https://github.com/psd401/aistudio/pull/1353) merged **35 seconds** before #1351 and removed `WS.fetchBrokerToken`, which the new skill called. Neither PR's CI could catch it — each ran against a dev that didn't yet contain the other, and the agent image is never built in CI.

**Restoring the import would have been the wrong fix.** #1353 established two rules the skill violated:

1. The model runtime never holds a reusable Google token. The skill minted one into the container.
2. The owner comes from a proxy-signed invocation context, never from the model. The skill took `--user` from argv.

It also targeted `/api/agent/workspace-token`, which #1353 retired to a 410 tombstone: *"Raw Workspace tokens are not available."*

So the lookup moves to where the token already lives.

## What changed

**New route** `app/api/agent/directory-lookup/route.ts` — owner from `verifyAgentInvocationContext` (never the body), agent token minted via `mintAgentWorkspaceTokenViaBoundary`, People API called server-side, and **only a shaped person record** returned. Body validation rejects unknown keys and requires exactly one of `email`/`chatId`, mirroring `workspace-execute`'s `isCommand()`.

**New lib** `lib/agent-workspace/directory-lookup.ts` — the People API logic, ported with every correctness property review found on #1351 intact:

- exact-address matching across **all** addresses, aliases included (`matchedAlias` reported)
- an empty-field `people.get` is a **miss**, not a nameless person
- 5xx classifies as `TRANSPORT` so callers retry
- a 2xx that won't parse **fails** instead of degrading to a false miss
- misses cached far shorter than hits

Cache is now in-process and bounded (500 entries), shared across invocations rather than per-container on disk.

**Skill rewritten** as a thin broker client — no `--user`, no token, no People API. Its remaining job is mapping the server's typed status onto the documented exit codes.

## Three allowlists had to agree

A missing entry in any one makes every call 404 at the relay:

| Layer | File |
|---|---|
| Skill side | `skills/_shared/agent-broker.js` |
| Python relay | `mantle_proxy.py` |
| The route | `app/api/agent/directory-lookup/route.ts` |

Tests assert the first two stay in sync, since nothing else fails loudly when they drift.

## Testing

- **24 jest tests** over the ported logic, including the no-enumeration property asserted against comment-stripped source
- **9 bun tests** on the skill: status→exit mapping, argv parsing, and posture assertions that it never reads a caller-supplied owner and never touches a token or `people.googleapis.com`
- **Proxy test** for the new relay route
- Fixed both eslint warnings in the file I added rather than leaving them for CI

## ⚠️ What is NOT verified

**The broker plumbing is untested end-to-end.** The route needs a proxy-signed invocation context that only exists inside a running agent container, so it cannot be exercised from a laptop. The People API behaviour underneath *was* live-probed extensively on #1351 with a real `agnt_` token — what's unproven is the relay → route → mint path. That needs a deploy to confirm.

Also unchanged and still yours: the directory now exposes student records domain-wide (Grade 3 name, `@edtools` address, grade, building), which is a separate admin decision noted on #1239.


krishagel:
@codex review — round 2. Both fixed in e752acf0.

**The P1 was mine to have caught, and I want to name why I missed it.** The cache was per-container before this PR, which partitioned it *by construction* — a container serves one owner. Moving it server-side into a process-global Map changed the sharing semantics, and I carried the keys across without carrying the partitioning. The bug was created by the move, not inherited.

You're right that it's a disclosure and not just staleness: directory visibility is per-account and this directory contains student records, so owner B could have received owner A's authorized result with B's token never used — and equally, A's visibility-based *miss* would propagate to everyone. Keys are now namespaced by the trusted owner from the signed context, and with no ownerKey the lookup skips caching entirely rather than falling back to a shared bucket.

**P2** — the broker helper already attaches the HTTP status, so `exitCodeForStatus` now uses it when no typed status is recognized: untyped 5xx maps to the documented retryable 12, untyped 4xx stays 2, and a typed status still wins (a 409 carrying DIRECTORY_SHARING_DISABLED keeps 16).

29 jest + 12 bun tests. The three isolation tests were verified to FAIL against the unpartitioned key (3 failed / 26 passed) rather than merely passing alongside the fix. Also fixed the eslint complexity warning the change introduced instead of leaving it for CI.

Still true and worth repeating: **the broker plumbing is unverified end-to-end** — the route needs a proxy-signed invocation context that only exists inside a running container.

krishagel:
@codex review — round 3. Both fixed in 0de05325.

**The eager-mint one is a regression I created, and it's the second time.** The container implementation had exactly this bug, you caught it on #1351 round 3, I fixed it there with a lazy token provider — and then dropped the laziness when porting the logic server-side. To make it harder to repeat, the parameter is now an explicit `TokenSource = string | (() => Promise<string>)` so the shape lives in the type instead of resting on a convention I can forget. The route passes a memoized minter and the token resolves only after a cache miss.

Your framing was sharper than a pure cost argument: the real damage is that a cached lookup *failed* whenever the mint boundary was down despite a valid answer being in hand. There's now a test for exactly that — a broken minter that throws, with the cached answer still returned.

**The HTTP-fallback one is a defect in my previous fix.** I put the fallback in the `default` branch, which is also where typed-but-unrecognized statuses land. Since the route returns LOOKUP_FAILED and FORBIDDEN with HTTP 502, a People API 400 was being advertised as a retryable outage. The fallback now applies only when no typed status is present at all.

32 jest + 13 bun tests. Both groups verified to fail against the pre-fix code (2 failed / 30 passed for eager minting; 1 failed / 12 passed for the precedence check).

Unchanged caveat: the broker plumbing is still unverified end-to-end — the route needs a proxy-signed invocation context that only exists inside a running container.

krishagel:
@codex review — round 4. Fixed in 20cfe712, and this was the best catch of the set.

I found three gates and updated them — the skill helper, the relay, and the route. **I did not know the Cedar policy existed.** Its comment is explicit that the omission is unrecoverable by convention: *"Fixed signed brokers. No wildcard below /api/agent/ is intentional."* So both apex and wildcard-subdomain entries are required, and `/api/agent/directory-lookup` had neither.

What makes this the worst-shaped gap of the ones you've found: the three gates fail differently. Miss the helper and it throws locally. Miss the relay and it 404s. Miss Cedar and the request is blocked **only in the deployed runtime** — every unit test, every CI job, and the entire local toolchain stay green while every production lookup fails. I could not have found this by testing.

The two-line Cedar fix is the small part. I also added `tests/unit/agent-broker-route-allowlists.test.ts`, which parses all three lists and asserts they agree — both directions between helper and relay, and Cedar coverage in both URL forms for **every** route, not just mine. It has a guard assertion that each parser found a non-trivial list, so a regex matching nothing can't make the rest vacuously pass. It lives in `tests/unit` (jest, runs in CI) rather than `infra/test`, which CI doesn't run.

Verified by deleting the Cedar entries and re-running: 2 failed / 3 passed, naming exactly the missing apex and wildcard forms.

Unchanged caveat: the broker path is still unverified end-to-end from here — it needs a proxy-signed invocation context that only exists in a running container.

krishagel:
@codex review — round 5. Both fixed in d6e9e471.

**The P2 is galling and worth admitting plainly.** I applied exactly this lesson twice earlier in this work — the psd-directory no-enumeration test and the migration-140 delete assertion both run against comment-stripped source for precisely this reason — and then didn't apply it to the drift test I wrote *about* allowlist drift. You were right that the helper and proxy parsers had the same weakness, so all three now strip first.

Two self-inflicted wounds getting there, both recorded in the commit:
1. My first stripper was a naive line-comment regex, which **destroyed every line in these files** — they're lists of URLs, and `https://` contains the marker. Took the suite from 5 passing to 3 failing. It's now quote-aware: walks each line and cuts only at a marker outside a quoted string.
2. The doc comment explaining that regex contained `*/`, which closed the block early and broke the parse.

Verified the way your finding describes — **commenting the Cedar entries out** rather than deleting them: 2 failed / 3 passed, restored 5/5.

**P3** — you're right that the server omits `cached` on a miss and only sets it true on a hit, so emitting verbatim dropped the documented field. `emit({ cached: false, ...result })` is back, with a test pinning the spread ORDER, since reversing it would make every cache hit report false.

15 bun + 5 jest allowlist tests green.

PR #1365, opened 2026-07-27
fix(db): bound DB waits and self-heal a wedged postgres.js pool

krishagel:
## Summary

Fixes the 2026-07-26 dev-server wedge: after concurrent load (a second E2E suite sharing the same local postgres), a `bun run server.ts` dev server entered a terminal state where **every DB-bound request hung forever** (0 bytes) while `/api/healthz` stayed healthy, the process idled at 0% CPU, and `pg_stat_activity` showed **zero** app connections. Requests were queued in the postgres.js pool behind capacity that never returned, and nothing in the stack applied a timeout — the process could not self-heal. Client aborts (Playwright's 60s timeout) never propagate into route handlers, so each aborted request also left a zombie awaiting the pool.

## Investigation (live reproduction)

On a dev server at this branch's base (`:3131`, Next 16.2.9 dev under Bun):

- An `ACCESS EXCLUSIVE` lock on `ai_models` + **100 client-aborted requests** to `/api/models` pegged the pool at 20/20 with an unbounded in-process queue; a fresh request hung to client timeout with **no server-side bound anywhere**. All 100 aborted requests still ran their queries to completion (zero abort propagation).
- Clean failure modes (lock release, `pg_terminate_backend` of all app backends) recovered — `executeWithRetry` handles them. The un-handled case is the incident's: capacity that never comes back and a queue with no deadline.
- 150 zombie handlers parked in a no-DB route left everything else healthy → the HTTP layer has no relevant cap; the DB pool queue is the choke point.

## Fix — three bounded layers (all env-configurable)

| Layer | Env | Default | Bounds |
|---|---|---|---|
| PostgreSQL `statement_timeout` | `DB_STATEMENT_TIMEOUT_MS` | 60s dev / off prod | any single statement, incl. lock waits |
| App-side pool deadline (`lib/db/pool-guard.ts`) | `DB_QUERY_DEADLINE_MS` / `DB_TX_DEADLINE_MS` | 90s / 300s | `executeQuery`/`executeTransaction` incl. **time queued in the pool** (which `statement_timeout` can't reach) |
| Wedged-pool self-heal | — | 3 consecutive deadline failures | discards + lazily rebuilds the client; `end({timeout:5})` force-rejects zombie queries; rate-limited to 1/60s |

Details that matter:

- `DbPoolDeadlineError` is deliberately **non-retryable** under `rds-error-handler` classification (retrying re-queues into the same starved pool).
- Drizzle builders are lazy thenables that **re-execute per `.then()` subscription** — the guard normalizes to a real promise so the race + detached-rejection swallow share one subscription.
- Real DB errors reset the wedge detector (they prove the pool is alive); only uninterrupted runs of deadline failures trigger a rebuild.
- Connections now set `application_name` (`aistudio`, override `DB_APPLICATION_NAME`) — during the incident there was no way to attribute connections in `pg_stat_activity`.

## Verification

- **Before:** wedge-analog (indefinite lock + aborted hammer) → fresh requests hang indefinitely.
- **After (tightened timeouts):** fresh requests fail bounded (`500` in ~statement-timeout), zombie backlog drains itself (pool sockets 21 → 1 in ~35s), healthz unaffected, **200 within 5s of the blocker clearing**.
- **After (default settings):** 100 aborted requests against a lock-blocked route → full recovery, `200` in ~15 ms immediately after release.
- `tests/unit/lib/db/pool-guard.test.ts`: deadline race, non-retryability, wedge threshold/reset semantics, single-subscription guard, env parsing (130/130 db unit tests pass).
- Full `bun run lint` + `bun run typecheck` pass (2 pre-existing lint errors were from my temp repro route, deleted).

## E2E gate status (intentional bypass, evidence below)

The pre-push e2e-local gate is currently **red on `dev` itself**: 34 functional specs fail identically on pristine `origin/dev` (b4f8a021) and on this branch — byte-identical failure lists (`diff` empty), 251 passed both runs. The failures cluster in the just-merged unified-content/agents + decision-graph areas (e.g. `/api/agent/consent-link` guard now returns 403 where the spec expects 401) — these functional specs are session-mint gated and never run in CI, so they didn't block those merges. Full authed suite was run **three times** for this PR (branch ×2, baseline ×1); the push used `SKIP_E2E=1` only after parity was proven. The upstream failures need their own fix.

## Notes / limitations

- The exact terminal state (zero sockets + dead queue) could not be re-created synthetically post-hoc — its incident logs were overwritten by the log-truncating restart (since fixed by port-scoped logs in #1355). The rebuild path is therefore covered by unit tests + construction, not a live end-to-end reproduction; the deadline layers were verified live.
- `/api/auth/session` hanging in the incident is not explained by the DB pool (its callbacks are pure JWT); the terminal state likely also involved Bun's socket layer. These bounds make DB-path starvation self-healing regardless of trigger.
- Prod `statement_timeout` stays **off** by default (no prod query-behavior change without explicit opt-in); the pool deadline defaults on everywhere since it only fires at ≥90s of queue starvation.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -175,6 +175,7 @@ mcp__awslabs_postgres-mcp-server__run_query
 - Use `DATABASE_URL` for local dev (set in .env.local)
 - Use `DB_HOST/DB_USER/DB_PASSWORD` for ECS (auto-injected from Secrets Manager)
 - Connection pool auto-manages connections (max: 20 per container)
+- Bounded DB waits + wedged-pool self-heal: `DB_STATEMENT_TIMEOUT_MS` (60s dev / off prod), `DB_QUERY_DEADLINE_MS` (90s), `DB_TX_DEADLINE_MS` (300s); per-call `deadlineMs` override — see `docs/database/drizzle-patterns.md` (Timeouts section)
 - Graceful shutdown: Handled automatically via `instrumentation.ts`
 - Connection warmup: Pools are pre-initialized on server startup
 
--- a/docs/database/drizzle-patterns.md
+++ b/docs/database/drizzle-patterns.md
@@ -952,6 +952,42 @@ async function healthCheck() {
 }
 ```
 
+### Timeouts, Pool Deadlines, and Wedged-Pool Self-Heal
+
+Added after the 2026-07-26 dev-server wedge: client-aborted requests (e.g.
+Playwright's 60s timeout) do not propagate into route handlers, so every abort
+left a zombie awaiting the pool, and nothing in the stack applied a timeout —
+a starved pool queue meant every DB-bound request hung forever with no
+self-healing.
+
+Three layers now bound DB waits (all configured via env, all in
+`lib/db/drizzle-client.ts` and `lib/db/pool-guard.ts`):
+
+| Env var | Default | What it bounds |
+|---------|---------|----------------|
+| `DB_STATEMENT_TIMEOUT_MS` | 60000 dev / 0 (off) prod | Server-side per-statement cap, including lock waits. A blocked query is cancelled by PostgreSQL (error `57014`, retryable). |
+| `DB_QUERY_DEADLINE_MS` | 90000 | Per-attempt app-side cap on `executeQuery` work, **including time queued in the postgres.js pool**. Fires only when work never reached a connection (it is above the statement timeout). Rejects with non-retryable `DbPoolDeadlineError`. |
+| `DB_TX_DEADLINE_MS` | 300000 | Same, for `executeTransaction` (higher: a transaction holds a connection across many statements). |
+
+`0` disables any of them. Per-call override for known-long operations:
+
+```typescript
+await executeQuery((db) => db.execute(longRunningSql), "bulkBackfill", {
+  deadlineMs: 0, // or a higher bound in ms
+});
+```
+
+**Self-heal:** three consecutive deadline failures with no other DB outcome in
+between mean the pool itself is wedged (queued work never reaches a
+connection). The pool is then discarded and rebuilt (`rebuildWedgedPool`),
+force-closing the old client after 5s — which also rejects every zombie query
+still queued on it. Rebuilds are rate-limited to one per 60s. A real database
+error resets the detector, since it proves the pool is alive.
+
+Connections also carry `application_name` (default `aistudio`, override via
+`DB_APPLICATION_NAME`) so the app's connections are attributable in
+`pg_stat_activity` during incident diagnosis.
+
 ---
 
 ## Related Documentation
--- a/docs/database/drizzle-patterns.md
+++ b/docs/database/drizzle-patterns.md
@@ -984,6 +984,22 @@ force-closing the old client after 5s — which also rejects every zombie query
 still queued on it. Rebuilds are rate-limited to one per 60s. A real database
 error resets the detector, since it proves the pool is alive.
 
+**Caveats:**
+
+- A deadline error does **not** cancel the underlying work — the query or
+  transaction keeps running until `statement_timeout` cancels it or a pool
+  rebuild destroys its connection. In production (where `statement_timeout`
+  defaults to off) a transaction can therefore still **commit after** the
+  caller received `DbPoolDeadlineError`. Do not build user-facing "retry on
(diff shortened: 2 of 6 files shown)


krishagel:
Dispositions for the review (addressed in 02d92096):

1. **Prod invariant / detector misfire** — documented rather than adding machinery, deliberately. A misfired rebuild's blast radius is bounded by design: in-flight statements fail with *retryable* connection errors (auto-retried by `executeWithRetry`), open transactions roll back atomically, and rebuilds are capped at 1/60s — one burst of connection churn, not a full outage. Corroborating with live pool state isn't feasible without reaching into postgres.js private internals. `drizzle-patterns.md` now spells out the prod interplay and prescribes per-call `deadlineMs` or `DB_STATEMENT_TIMEOUT_MS` for known >90s prod operations. Enabling a prod statement_timeout default is a real behavior change we'd rather make as its own decision.
2. **Commit-after-deadline** — documented in `drizzle-patterns.md` (idempotency-keys callout for retry flows built on `executeTransaction`).
3. **`rebuildWedgedPool` coverage** — direct unit test added (`rebuild-wedged-pool.test.ts`): discard + `end({timeout:5})`, lazy rebuild, cooldown suppression. The function is now exported for tests/ops only.
4. **Numeric `statement_timeout`** — verified live before the PR: with `DB_STATEMENT_TIMEOUT_MS=5000`, lock-blocked queries were cancelled at ~5s on a real dev server (the wedge repro run), so the numeric startup parameter lands.
5. **Semicolons** — left as-is; `lib/db` is already mixed (`rds-error-handler.ts` is semicolon-free) and no lint rule enforces it.

krishagel:
🤖 **pr-fix routine: nothing to change — deferring to your decision**

Picked this PR up because two unresolved P1 review threads were open. On inspection, you had already responded to both in `02d9209` (7 minutes after the review landed), so I made no code changes.

**Current state**
- CI: **6/6 green** on `02d9209` (Test/Lint/Typecheck, CDK validate, Unified Content lifecycle, CodeQL ×2, claude-review).
- `mergeable_state: clean`, no `CHANGES_REQUESTED`.
- Review comments: 2 — both from `chatgpt-codex-connector`, both addressed in `02d9209`.
- Failing checks: 0.

**Why I stopped rather than "fixing" the two P1s**

Both threads are still shown as unresolved only because bot authors never resolve their own threads. Your follow-up commit addressed them as **deliberate, documented accepted risks** rather than as code changes, and I'm not going to autonomously overrule that:

| Thread | Reviewer asked for | `02d9209` did |
|---|---|---|
| [r3654234008](https://github.com/psd401/aistudio/pull/1365#discussion_r3654234008) — cancel timed-out work | Cancel/prevent the operation before returning the error | Documented that a deadline does not cancel the work, and that callers must not build user-facing retry flows on `executeTransaction` without idempotency keys |
| [r3654234012](https://github.com/psd401/aistudio/pull/1365#discussion_r3654234012) — distinguish pool waits from active work | Only confirmed queue waits should feed the wedge detector | Documented that with `statement_timeout` off the two are indistinguishable, and that a misfired rebuild is *deliberately survivable* (retryable connection errors, atomic rollback, 1-per-60s cap) |

That's a legitimate call, and the reasoning in the commit message is sound. Flagging it explicitly only so the accepted risk is a conscious choice rather than something that quietly rode along: as written, three slow-but-healthy operations in a row can still force-close a live pool in prod, and a mutating `executeQuery`/`executeTransaction` can still commit after the caller saw `DbPoolDeadlineError`.

**If you'd rather enforce those in code**, both are implementable without a prod `statement_timeout`:
1. Give `DbPoolDeadlineError` a `phase` (`queued` vs `active`) and count only `queued` toward the wedge detector — plus corroborate with a `select 1` liveness probe before rebuilding, so a pool that still answers is never torn down.
2. Race an `active`-phase deadline *inside* the `db.transaction()` callback so an expired deadline rejects the callback and the driver issues `ROLLBACK` — which makes "error ⇒ nothing committed" true again for `executeTransaction`. (The outer deadline needs a few seconds of headroom so the rollback round-trip can finish first.)

Say the word and the routine will implement that on the next pass.

**Label**: adding `pr-fix-stuck` so the routine stops re-picking this PR every ~3h while it's effectively resolved. Remove the label if you want it re-evaluated, or just merge.

---

PR #1384, opened 2026-07-27
fix(agent): SigV4 Bedrock, scheduled-task deadline + job promotion, 70s cold start, and revive psd-freshservice

krishagel:
Five fixes to the agent platform, all traced by measurement rather than inspection. Three were already on this branch and are deployed to dev; two are new.

## 1. Bedrock via execution-role SigV4 — no bearer key, no proxy hop

The model-facing container authenticated to Bedrock with a bearer key hydrated from Secrets Manager and **inlined into `openclaw.json`**, inside the same container that runs 33 model-authored skills. #1353 replaced that with a loopback proxy to a web broker, which removed the on-disk key but put an authenticated ALB hop in front of every model call.

SigV4 from the AgentCore execution role removes both. The evidence this works from inside a microVM: the `memorySearch` path has used exactly this mechanism in production since #1184, just scoped to Titan embeddings.

`us.anthropic.claude-sonnet-5` is a **cross-region inference profile**, so the grant names the profile ARN *and* the foundation-model ARN in all three regions it routes to. Profile-only fails 100%; a missing member region fails *intermittently*, since routing is per-request.

**Deployed and verified:** `BOOT_OK provider=amazon-bedrock`, turns completing in 8–15s.

## 2. Scheduled tasks: derive the deadline from real remaining time

The Morning Dispatch reported "Agent temporarily unavailable" on 2026-07-27. It had not failed — the agent finished, **5.6 seconds after the Lambda hung up on it**.

The old stack assumed the turn starts when the request is sent. It starts once AgentCore has cold-started the microVM, and that gap is real. The deadline is now derived from the Lambda's actual remaining time.

The user-facing message for a timeout is also its own now: "temporarily unavailable, try again later" was not merely unhelpful, it was **false** — and it sent the investigation looking for an outage that never happened.

## 3. Promote long scheduled tasks to the background job-runner

Lambda's 15-minute ceiling is an AWS hard limit, so a scheduled task that legitimately needs longer could never finish. The router has solved this for interactive turns since #1138; cron was never wired up. Same cluster, same task definition, same payload contract.

Only reachable *because* of #2: under the old fixed deadline the harness never got to report `ChatDeadlineExpired`, so promotion could not have triggered. A test pins that `OpenClawChatError` is **not** promotable.

New run status `promoted` — neither success (nothing produced yet) nor error (nothing broke). No migration needed: `agent_scheduled_runs.status` is `VARCHAR(32)` with no CHECK constraint.

## 4. Cold start was the workspace restore, not the model

Measured on the deployed dev runtime:

```
+  0.0s  BUILD_MARKER          microVM starts
+  9.9s  BOOT_OK               gateway ready
+ 50.3s  workspace pull        files=1100 elapsed_s=70.8   <-- the wall
+ 65.6s  Invocation complete   turn itself: ~15s
```

Boot is 10s, the model turn is 15s. Of 1,100 files restored, **812 are session transcripts totalling 202 MB — 93% of the bytes**. They are archive, not memory: interactive session keys embed the image digest (every deploy makes prior transcripts unresumable) and scheduled keys embed the date. Continuity lives in `memory/`, 55 files, still restored in full.

`ListObjectsV2` already returned size and mtime and they were being discarded, so exposing them costs **no extra S3 call**. `paths` is retained — containers deploy independently of the web tier, so an older image is always in flight during a rollout.

**Safe by construction:** this skips a *download*, never an upload or delete. `push_workspace` only walks local files; nothing enumerates S3 to remove keys. A test asserts that invariant directly, because if it stopped holding, an unrestored file would become a *deleted* one — the failed-READ-becomes-destructive-WRITE shape that destroyed agent memory on 2026-07-27.

## 5. psd-freshservice has been dead since #1353

`lib/api.js` exec'd `path.resolve(__dirname,'..','..','psd-credentials','get.js')`. #1353 removed plaintext credential access and deleted that script; this skill was never migrated. All 13 commands failed **before** checking whether a key was provisioned, so the error read "your Freshservice credential is missing" — blaming the user for a broken skill.

Found only because the path is built with `path.resolve`; a literal grep matches nothing. Blast radius checked: `psd-redrover` was migrated at the time, `psd-image-gen` uses `check_capability.js` (still exists). This was the only one left behind.

The key now stays server-side. **The (method, path) allowlist is the security boundary** — the model composes the path, so without it this is a general-purpose proxy signing arbitrary requests with the owner's key. Query is split off *before* matching so every pattern stays a static anchored literal; folding an optional query group in produces nested quantifiers over an attacker-controlled string (a real ReDoS hazard, and a lint error I hit and fixed). The query charset excludes `/`, so a query cannot smuggle path segments.

No allowlist changes needed — `/api/agent/credentials` is already in all three gates.

---

## Verification

- **3,385 unit tests** pass; 35 python tests pass; typecheck clean.
- Lint at the **497-warning / 0-error** baseline — including removing the ReDoS error and two warnings my own code introduced.
- `cdk synth` renders the Bedrock grant, and all six cron `JOB_*`/lock env vars with `PassRole` on both task roles.
- **Mutation-tested where it matters:** removing only `obtainExecutionRole().roleArn` from the cron PassRole grant — the subtle half, which fails at launch with an opaque error — fails exactly one test and no others.

### Two things reviewers should know

**E2E was pushed with `SKIP_E2E=1`, deliberately.** The local suite runs 285 passed / 1 failed. The failure is `atrium-meridian-editor.spec.ts` (bold button never sets `aria-pressed`). I verified it fails **identically on clean `origin/dev`** with none of these commits, and this branch touches zero Atrium/editor files. It is a pre-existing dev breakage, not a regression here — but it is real and currently red on dev.

**One flaky unit failure.** A single run showed 1 failed suite; it did not reproduce across three subsequent full runs and I could not capture which suite it was. Recording it rather than claiming a clean sweep.

## Deploy order

Web tier is safe to deploy first: `paths` still works for running containers, and the freshservice fix is inert until the image ships. Then image build for `workspace_sync` + the skill.


krishagel:
**Re: Codex P1 — usage accounting when bypassing the proxy**

Real problem, but not a regression from this PR: token accounting is already reporting zero on **both** paths.

```
dev  (after SigV4 switch):  tokens_in=0 tokens_out=0 tool_calls=19 model_calls=20
prod (old path, 24h):       tokens_in=0 tokens_out=0   (every invocation)
```

Prod runs `2026-07-23-prod-cutover`, which never used the loopback proxy for chat, and it reports zero as well. So the proxy was not successfully extracting usage before this change either — `_extract_usage` / `inject_include_usage` in `mantle_proxy.py` are not landing counts on `Invocation complete`.

Note the counters that *do* work are unaffected: `tool_calls`, `model_calls`, `elapsed_s` and `duration_ms` are emitted by the wrapper, not the proxy.

Merging this as-is rather than widening scope — restoring real token accounting is its own fix, and it needs to work on the SigV4 path regardless of what this PR does. Filing it separately.

The `mantle_proxy.py` process itself stays: it still relays every `/agent-broker/api/agent/*` route, so nothing is orphaned by the chat path moving off it.

PR #1386, opened 2026-07-27
fix(infra): db-init asset hash + compile raw-asset Lambdas at synth time (no more committed compiled twins)

krishagel:
## Context

A triage task flagged that the committed compiled artifact `infra/database/lambda/db-init-handler.js` might be stale relative to its `.ts` source (missing the CREATE TYPE / SQLSTATE 42710 idempotency guard), and asked whether CDK packages that committed `.js` into the deployed db-init Lambda.

## Verification of the original claim

- **The committed `.js` was not what deploys.** DatabaseStack bundles at synth time (`npm install && npm run build`, ships `dist/*`), so deployed code always compiles from the `.ts`.
- **The committed `.js` was current anyway** (regenerated in f7cd0dac), and **deployed environments do have the 42710 guard** — its PR (#1061) also changed a schema file already covered by the custom asset hash, so the asset was rebuilt on the next deploy.

## Commit 1 — the real defects behind the report

**db-init asset hash omitted the Lambda sources.** `assetHashType: CUSTOM` fully replaces CDK's default source hash, and the custom hash covered only `migrations.json` + `schema/`. A handler-only change kept the old hash → CDK silently reused the stale asset → the fix never deployed. The 42710 guard only shipped by luck. Fixed: the hash now covers the lambda sources, `package.json`, and `tsconfig.json`, extracted into `computeDbInitAssetHash()` with 9 unit tests pinning the input surface (`infra/test/unit/db-init-asset-hash.test.ts`).

**A sibling twin was genuinely stale — and deployed.** `file-processor`/`url-processor`/`textract-processor` use raw `Code.fromAsset`, so their committed `.js` twins WERE the deployed code. `textract-processor/index.js` predated the `chunkText` lineStart fix from 6014d13c: deployed chunk `metadata.lineStart` was always 0 for ~2.5 weeks (the deferred REV-INFRA-123 item from the B-013 scope notes).

## Commit 2 — eliminate the twin mechanism entirely

Rather than regenerating twins, make staleness structurally impossible:

- **`bundledLambdaAsset()` in processing-stack** compiles `file-processor`, `url-processor`, and `textract-processor` from `.ts` at synth time — same local-bun/Docker-npm split and `AssetHashType.SOURCE` as the existing agent-router asset, with `node_modules`/`dist`/`__tests__`/twins excluded so the source hash is deterministic across machines. Production `node_modules` are vendored into each bundle so every Lambda runs its own pinned deps — **file-processor needs pdf-parse 1.x**, and the shared processing layer's old build line installed pdf-parse *unpinned*, which now resolves to the incompatible 2.x API.
- **secret-cache layer** compiles at synth time too, preserving the flat `/opt/index.js` layout (runtime provides the AWS SDK; nothing vendored).
- **All committed compiled twins deleted** (file-processor ×3, url-processor, textract-processor, secret-cache, and the db-init pair — which were never packaged but misled this investigation into existing) and their paths gitignored so they can't come back.
- **Each lambda dir pins its own toolchain** (`typescript ^5`, `@types/*`, `build` script) so Docker-fallback bundling is self-sufficient — previously textract type-resolved by accident from `infra/node_modules`, which fails under the repo's TS 6.0.3.
- **build-lambdas.sh repaired**: the file/url compile+install steps were doubly broken (TS5112 under TS 6.0.3 aborts the script at line one; compile-before-install resolved infra's pdf-parse 2.x on fresh checkouts) and are now obsolete — removed. The processing-layer step now delegates to `layers/processing/build-layer.sh`, respecting the committed pinned manifest instead of installing packages by name (the old line would have silently rewritten the B-013 `node-fetch ^2.7.0` pin with node-fetch 3.x). agent-router step unchanged.
- **url-processor artifact tests** now assert on `index.ts`: with synth-time compilation the source *is* the deployed artifact; there is no twin left to drift.

## Testing

- All four converted Lambdas build standalone (`bun install && bunx tsc` → `dist/`)
- infra `bunx tsc` clean; jest `lambdas` project 197/197; jest `test/unit` 179 passing (`agent-chat-bridge` has a pre-existing isolation coupling — synth-time `bun install` in any lambda source dir splits the `@aws-sdk` instance its test mocks; reproduced on untouched code, passes on a clean tree)
- Full CI-style `cdk synth --all --no-lookups` (CI's dummy account env) succeeds across all 29 stacks
- Staged assets verified: file-processor bundle contains compiled JS + vendored **pdf-parse 1.1.4**; textract bundle contains the lineStart fix; layer asset keeps flat `index.js`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


krishagel:
@codex review

Both review findings are addressed in 90285894:
- **Codex P2 (nested sources)**: `collectLambdaSourceFiles()` now walks the Lambda dir recursively (skipping `node_modules`/`dist`) and hashes each file with its relative path, so nested sources and renames both invalidate the asset hash. Regression tests added.
- **CodeQL file-system race (570/571)**: the `statSync`-then-`readFileSync` pattern is gone — directory entries are filtered via `readdirSync(..., { withFileTypes: true })` dirents, no separate stat calls.

Note: commits 4c615345+ extend the PR beyond what Codex reviewed at 61bfe30d — the raw-asset Lambdas (file/url/textract processors, secret-cache layer) now compile at CDK synth time and their committed compiled twins are deleted; please review the full diff.

krishagel:
@codex review

Addressed the dependency-drift P2 in 338a0436: `bun.lock` is now committed for all three processor Lambdas and the secret-cache layer, included in the SOURCE asset hash (a lockfile change invalidates the asset), and installed with `--frozen-lockfile` in the primary local bundling path for both the build install and the vendored production install. Dependency updates now deploy via lockfile changes instead of drifting silently. The Docker fallback remains npm-based (no bun in the bundling image — same accepted divergence as the agent-router asset; CI uses the local path).

PR #1389, opened 2026-07-27
fix(agent): restart a scheduled task in a fresh session when context overflows

krishagel:
Closes the hole left by #1384: promotion fired only on `ChatDeadlineExpired`, but the prod Morning Dispatch fails with **context overflow**. That arrived as the generic `OpenClawChatError`, which is deliberately not promotable — so the one failure mode actually killing the dispatch was the one promotion could not catch.

## Why resuming isn't enough

The obvious fix — make `OpenClawChatError` promotable — is wrong twice over. It would hand every genuine crash a two-hour retry budget, **and it wouldn't work**: AgentCore sticky-routes by session id, so resuming hands the runner the exact transcript that outgrew the model window. It re-overflows on the first model call, having spent a Fargate cold start to get there.

Overflow needs the opposite treatment from a deadline:

| Reason | Meaning | Runner |
|---|---|---|
| `deadline` | healthy transcript, ran out of clock | **resume** — the in-progress work is the point |
| `context-overflow` | the transcript *is* the problem | **restart** in a fresh session |

## Changes

- **`harness_adapter.py`** — `_classify_chat_error()` names `ContextOverflow` separately. The classification lives where the message is produced, not as a regex duplicated into TypeScript. Unrecognized wording keeps the generic class, so a message change degrades to today's behaviour rather than misclassifying a crash as recoverable.
- **`job-promotion.ts`** (both bundles) — `promotionReason()` returns `'deadline' | 'context-overflow' | null`, carried on the payload. An absent reason parses as `deadline`, so a payload from an older cron build still resumes.
- **`job-main.ts`** — on overflow, invokes a *derived* session id with a restart prompt. The lock stays on the **original** session id — that's what cron acquired and what the router checks for "still working on your earlier task".
- Session ids suffix once (`-r2`, `-r3`), not cumulatively — AgentCore enforces a length limit.
- Cron sends a different ack: "I've moved it to a background job" overstates what carries over when the previous attempt is being discarded.

The restart prompt is deliberately **not** the continuation prompt. "Continue from where you stopped" is meaningless with no history and invites the model to hunt for work it can't see. It leads with the real hazard — the previous leg's tool calls aren't visible, so side effects may already have run.

## Deliberate divergence from the router

Scheduled tasks restart on overflow; **interactive turns do not**. A restart discards the user's conversation, and an interactive user is right there — they can `/reset` or rephrase, and they can see something went wrong. Nobody is watching a 6am dispatch, and its request is fully described by the stored schedule prompt.

Pinned by a test rather than left implicit, because the existing cron/router parity test would otherwise have been quietly routed around.

## Verification

4,210 unit tests pass; 33 python tests; both Lambdas compile; typecheck clean; lint 0 errors (501 baseline unchanged, none in these files).

**Mutation-tested twice:**
- reverting the runner to `job.sessionId` fails exactly the "invokes with the RESTART session id" test and no other
- renaming the Python class constant fails exactly the cross-language contract test — the failure that would otherwise surface weeks later as the dispatch quietly breaking again

A cross-language test reads `CONTEXT_OVERFLOW_ERROR_CLASS` out of `harness_adapter.py` and asserts the TypeScript promotion keys off that literal. Nothing at build time links a Python string to a TypeScript comparison.

## Deploy

Needs **both** an image build (`harness_adapter`) and a web/Lambda deploy (cron + router).

Pushed with `SKIP_E2E=1`: the only local E2E failure is `atrium-meridian-editor.spec.ts`, which fails identically on clean `dev` and is unrelated.

### DIFF
--- a/infra/agent-image/harness_adapter.py
+++ b/infra/agent-image/harness_adapter.py
@@ -82,6 +82,37 @@ def _format_for_chat(text: str) -> str:
         return text
 
 
+CONTEXT_OVERFLOW_ERROR_CLASS = "ContextOverflow"
+
+
+def _classify_chat_error(error_message: str) -> str:
+    """Name the chat-error class from OpenClaw's message.
+
+    Every chat-channel error arrives as the same generic OpenClawChatError, with
+    the only distinguishing detail buried in free text. That conflates two very
+    different situations:
+
+      • Context overflow — the transcript outgrew the model's window. The work
+        itself is fine; the SESSION is the problem, and continuing it is
+        guaranteed to fail again. Recoverable, but only by starting fresh.
+      • Everything else — a genuine fault. Retrying is not obviously safe.
+
+    Downstream (agent-cron promotion, agent_failures, alarms) has to tell these
+    apart, and the classification belongs HERE, where the message is produced,
+    rather than as a regex duplicated into TypeScript. On 2026-07-27 the prod
+    Morning Dispatch hit overflow twice, burned ~7 minutes retrying, and its
+    failure was indistinguishable from a crash.
+
+    Matches on the stable part of OpenClaw's wording ("context overflow" /
+    "prompt too large"); an unrecognized message keeps the generic class, so a
+    wording change degrades to today's behaviour rather than misclassifying.
+    """
+    lowered = (error_message or "").lower()
+    if "context overflow" in lowered or "prompt too large" in lowered:
+        return CONTEXT_OVERFLOW_ERROR_CLASS
+    return "OpenClawChatError"
+
+
 def _frame_failed_partial(partial: str) -> str:
     """Wrap a failed/degraded turn so it is never presented as a clean answer.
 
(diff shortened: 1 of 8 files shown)


krishagel:
Both P1s fixed in `8192a127` — both were real, and both were mine.

**1. Truncated request on restart.** Correct and the more serious of the two: a restart has no transcript to supply the omitted half, and schedule validation accepts up to 20,000 chars against a 2,000-char excerpt cap. That failure *looks like success* — it posts a confident answer to the wrong question.

A restart now carries the prompt in full, and `buildJobPayload` enforces the 8 KiB RunTask override cap itself. When the full prompt can't fit it **throws** rather than quietly truncating; cron catches it, declines to promote, and posts the partial. A partial that admits it's partial beats a restart that confidently does the wrong work.

**2. Colliding restart session ids.** Also correct, including the sharpest part of the finding — that the `-r3` increment path my test covered *is unreachable in production*, since cron always passes the original date-based id. My test was green and testing nothing. Uniqueness now comes from the lock token (a fresh UUID per promotion), suffix replaced rather than appended and bounded to 8 hex chars for the session-id length limit. The test now asserts the real production shape: two promotions of the **same** date-based id must differ.

**Mutation-verified both ways:** restoring the counter suffix fails exactly the uniqueness test and no other; re-truncating the restart prompt fails exactly the two restart-prompt tests and no others.

4,215 unit tests pass, typecheck clean, both Lambdas compile, lint 0 errors. The cron/router builders remain byte-identical, so the duplicated payload logic hasn't drifted.

Issue #1284, opened 2026-07-22
[Atrium] Add immutable content-asset upload and render APIs for authored screenshots


krishagel:
## Problem statement

Atrium's design and S3 adapter reserve `atrium/objects/{objectId}/assets/{assetId}`, but the external content surface has no route, metadata model, or renderer contract for uploading and embedding an authored image. The unified repository ingestion work does not close this gap: repositories model source knowledge, while Atrium content objects remain the authored/published domain.

Atrium Capture needs to attach one or more locally redacted screenshots to a draft, reference them from immutable Markdown versions, render them for permission-checked internal readers, and keep the same asset pinned when that version is published. Presigned S3 read URLs cannot be stored directly in Markdown because they expire; arbitrary public image hosting would bypass Atrium visibility, lifecycle, and audit controls.

Current evidence on `dev` at `ba2cc4cae06fd49e5954d0419a407c21174e954c`:

- `lib/content/storage/s3-store.ts` implements `assetKey()` and hard-delete cleanup for an object's asset prefix.
- `docs/features/atrium-design-spec.md` reserves the same asset path.
- `app/api/v1/content/**` exposes object, version, visibility, publish, delete, and OKF routes, but no asset route.
- The rich-media URL gate currently accepts only absolute HTTP(S) URLs, so no durable object-scoped asset reference can be authored today.

## Proposed API and contract

Add an immutable `content_assets` record and a two-step upload flow. Reuse the existing documents bucket and Atrium object prefix; do not expose AWS credentials.

### Initiate upload

`POST /api/v1/content/{id}/assets`

Required scope: `content:update`; service must also enforce the same object edit gate as version creation.

```json
{
  "filename": "step-03.png",
  "contentType": "image/png",
  "byteLength": 483291,
  "sha256": "base64url-sha256",
  "purpose": "capture_step",
  "width": 1440,
  "height": 900
}
```

```json
{
  "data": {
    "id": "asset-uuid",
    "objectId": "content-uuid",
    "state": "pending",
    "upload": {
      "method": "PUT",
      "url": "https://presigned-upload-url",
      "headers": {
        "content-type": "image/png",
        "x-amz-checksum-sha256": "base64-sha256"
      },
      "expiresAt": "2026-07-22T22:00:00Z"
    },
    "embedRef": "::atrium-asset{id=\"asset-uuid\" alt=\"Step 3\"}"
  },
  "meta": { "requestId": "req_..." }
}
```

### Complete upload

[...]

PR #1399, opened 2026-07-27
feat: psd-sop-creator agent skill + Atrium asset/source broker routes

krishagel:
## Summary

Adds **`psd-sop-creator`**, an agent skill that turns a procedure — described in chat, or carried over from a PDF, a Google Doc, or an existing Atrium document — into a Peninsula School District **Standard Operating Procedure**, filed in Atrium as an **editable markdown document** (`kind=document`, not an HTML artifact), **private**, in **draft** status, in a dedicated collection.

An SOP is a living document staff revise; an artifact would be a snapshot nobody can edit. The skill never publishes and never widens visibility — review stays a human step, and `draft` is the extension point if an approval workflow is ever added.

Getting there required three platform gaps to be closed first, all in this PR.

## Changes

**1. Agent Atrium broker — asset and source routes** (`app/api/agent/atrium/route.ts`, `lib/agent-workspace/atrium-owner-operation.ts`)

The broker could create and version content but could neither *read* a document's body nor attach an image to one:

- `GET /<id>` deliberately omits a document's text (`bodyLocation: "proof"`), so an agent could write documents it could never read back. Added `GET /<id>/source`.
- There was no `/assets` route, so the canonical `::atrium-asset` embed (#1284) was unreachable. Added `POST /<id>/assets`, `POST /<id>/assets/<assetId>/complete`, `GET /<id>/assets`, and a bounded `GET /<id>/assets/<assetId>/bytes`.

The initiate schema mirrors the v1 REST contract exactly (PNG/JPEG/WebP, 20 MiB, base64url SHA-256), so the agent path cannot reserve an upload the human path would refuse. The byte read is capped at 4 MiB — checked against both the declared length and the real post-normalization length — because it crosses the loopback broker base64-encoded in a JSON envelope. Assets are resolved through the *object* first, so an asset id belonging to another object 404s rather than being served under this object's path.

**2. `psd-atrium` — `read-source`, `upload-asset`, `get-asset`, `list-assets`**

`upload-asset` runs reserve → presigned PUT → complete in one command. Content type comes from **magic bytes**, not the filename (the server re-derives it, so a `.png`-named JPEG would fail completion with an opaque rejection). Bytes go straight to S3 rather than through the broker; the URL is short-lived, checksum-pinned, comes from the Atrium response itself, and must be `https`. `get-asset` exists because assets are per-object — reusing an image means copying it, not re-referencing it.

**3. `psd-pdf-to-markdown` — `--extract-images`, plus a stdout fix**

Opt-in image extraction with absolutized references and a 50-image cap. Default behavior is byte-for-byte unchanged.

Also fixes a **pre-existing defect**: PyMuPDF prints parser/OCR progress to **stdout**, breaking this script's documented JSON-only contract. `contextlib.redirect_stdout` does *not* fix it — the noise comes from the extension module writing to fd 1 directly — so the fix dups fd 2 over fd 1 around the conversion.

**4. Migration 161** — seeds the `standard-operating-procedures` collection with `default_visibility_level = 'private'`, so an agent-drafted SOP can never default to district-wide visibility. Verified by execution against real PostgreSQL 16: the row lands with the right slug, `private` visibility, position 5, `nav_item_id` NULL — and re-running it is a true no-op (still exactly one row), confirming the `NOT EXISTS` guard.

**5. Branding** — the two-color district logo at `public/branding/`, downscaled 3001px → 300px (`.atrium-content img` is `max-width:100%`, so a 3001px logo would fill the entire document column).

## The template gate

Every rule in `validate` exists because breaking it loses content **silently** in Atrium — no error, the content simply is not there:

| Rule | Why |
|---|---|
| No raw HTML | `markdown-bridge.ts` overrides marked's `html` renderer; a `<table>` is dropped at parse time |
| No `data:` URIs | The sanitizer strips them |
| Images on their own line | An inline image splits the paragraph into three blocks, cutting the sentence in half |
| No author-supplied H1 | The skill injects the letterhead and heading; they would be duplicated |

Section order is enforced as a **subsequence of one canonical list**, so required order and optional insert positions come from a single rule that cannot disagree with itself. Checks run off a fence-aware scan, so a ` ```html ` example is never mistaken for markup. Violations exit **3** with a structured list, each carrying a `fix` — a to-do list, not a dead end.

## Verification

**Spikes against the real collab bridge, before designing anything:** markdown tables, image lines, and `::atrium-asset` directives all survive `seedYDocFromMarkdown` into the TipTap schema — so the metadata block is a real table, not the bold-label fallback. An inline image was confirmed to split its paragraph, which is why that is a violation rather than a style note.

**End-to-end, through the real `spawnSync` path** against a fake `psd-atrium`, then feeding the produced document back through `seedYDocFromMarkdown`:

```
image: https://…/branding/psd-logo-2color-horizontal.png [alt=Peninsula School District]
heading(h1): Standard Operating Procedure (SOP)
table(5 rows)
heading(h2): Title … Scope … Procedure … Safety Considerations …
             Quality Control … References … Revision History
IMAGE NODE: {"src":"/api/v1/content/assets/<uuid>/bytes","alt":"Control panel","assetId":"<uuid>"}
paragraph still holding a raw directive? false
```

**PDF extraction, against the real "Printer & Copier Allocation" SOP PDF:** 3 images extracted with absolute references, stdout parses with `json.load`, progress noise on stderr, and the no-flag path still emits no image references.

- Build/typecheck: ✅ `tsc --noEmit` clean
- Lint: ✅ 0 errors. The two web-tier files I touched went from 3 pre-existing warnings to **zero**
- Tests: ✅ **4,273 passed** (full jest suite), plus 103 skill tests and the Atrium smoke suite
- Agent-image static gates: ✅ bootstrap budget + config self-consistency
- E2E: not applicable — no UI surface in this change (`e2e_required: false`)

## Evidence

_N/A — no UI surface._ This adds an agent skill, a broker route, a CLI flag, and a migration. The verification above is the executable evidence; the collab-seeding output is the closest thing to a rendering proof, and it is reproducible from the test suite.

## Notable cleanups pulled in

- **Decomposed `executeOwnerAtriumOperation`.** The new routes pushed it to complexity 58 across 283 lines. The position of the authoring-capability assert is the security-relevant line in that function, and with reads and writes interleaved nothing structural stopped a future read from landing *below* the gate — where it would demand authoring rights just to look at a document. Now: read → gate → content / metadata / asset / publish writes.
- **Fixed two brittle migration tests.** `rooms-migration` and `student-rooms-navigation-migration` asserted their position with `slice(-2)` / `slice(-3,-1)` — effectively "I am the last migration" — so they broke on the next unrelated migration anyone added, including this one's. They now assert the invariant that actually matters (relative order against the migration each depends on).
- **CI wiring.** `test:skill:sop-creator` **and** `test:skill:atrium`. psd-atrium's suite had never run in CI at all — plain CJS outside the jest roots — despite guarding every agent write into the content workspace.

## Review round 2 — five findings fixed

All five were real. Details in commit `359111e4`; summary:

1. **P1 — the skill could not perform its own advertised PDF flow.** SKILL.md told the agent to run `psd-pdf-to-markdown` (a Python entrypoint) while the frontmatter granted only `Bash(node:*)`, and run.js carried `pdf`/`workspace` runners nothing called. Fixed by **narrowing**: dead runners removed, and ingest is now documented as a separate skill invocation under *those* skills' permissions. Widening this skill's grant would have bought nothing — it only ever shells psd-atrium.
2. **P2 — a large SOP failed to spawn with `E2BIG`.** One argv value is capped at 128 KiB, far below the broker's 4 MiB. Added `--markdown-file` / `--body-file` to psd-atrium (mirroring the existing `--code-file`), and psd-sop-creator now *always* uses them — so the large-document path is exercised every run rather than being the one nobody tests.
3. **P2 — failed runs left orphan empty drafts.** Images can only upload after the object exists, so a typo'd path aborted having already persisted a bodyless draft, and every retry added another. Now: preflight every local image (`accessSync`, not `existsSync` — an unreadable file exists but still can't be uploaded) and catch missing `--source-id` *before* creating anything; if a later step fails, delete the object that was created.
4. **P2 — `collectImages` ignored code fences** while the validator honored them. A technical SOP documenting markdown would have its *example* image uploaded and the example rewritten into a live directive — corrupting the thing being documented.
5. **P2 — dropped PDF images left dangling references.** Past the 50-image cap the files were deleted but their `` links remained, returning `status: ok` for a broken document. The cap now applies before the Markdown is written and the orphaned references are removed.

**CodeQL:** fixed the one real finding (`js/insecure-temporary-file` — a test used a predictable `tmpdir()+Date.now()` path). Also hardened `get-asset` to verify decoded bytes are genuinely PNG/JPEG/WebP before writing, so it can never write arbitrary response bytes to a caller-named path; a negative control confirmed the new test fails without the guard.

> Note: `gemini-code-assist[bot]` has been sunset and `copilot-pull-request-reviewer[bot]` is quota-limited, so Codex is the only reviewer producing findings on this PR.

## Follow-up (post-deploy, manual)

- [ ] Deploy with the canonical command + a rebuilt `agentImageDigest` (this PR does **not** deploy; no new Dockerfile `RUN` layer is added, so the Firecracker layer ceiling is unchanged).
- [ ] Verify an agent turn end to end: build an SOP from the sample PDF, confirm the letterhead, metadata table, seven sections, and carried-over screenshots render in the Atrium editor, and that the document is private + draft.


krishagel:
## Review round 2 — all findings addressed

Every finding was real. Fixed in `359111e4` (+ `6aa3a636` for the doc follow-up), and all 8 checks are green at `6aa3a636`.

| # | Finding | Resolution |
|---|---------|-----------|
| P1 | `allowed-tools: Bash(node:*)` blocks the PDF converter the workflow calls for | Fixed by **narrowing**, not widening. The `pdf`/`workspace` runners in `run.js` were dead code — nothing called them — so they and the unused `VENV_PY` are gone. SKILL.md now states that ingest is a **separate skill invocation** the agent makes first, under *those* skills' own `allowed-tools`, handing the result to `create --body-file`. Granting this skill the Python entrypoint would have widened its permissions for no gain. |
| P2 | Large bodies fail to spawn with `E2BIG` | Added `--markdown-file` (create-document) and `--body-file` (edit) to psd-atrium, mirroring the existing `--code-file`, sharing one `readInlineOrFile` helper that rejects the both-flags case. psd-sop-creator now **always** uses them — so the large-document path runs every time rather than being the one nobody exercises. Covered by a test with a 200 KB body. |
| P2 | Orphan bodyless drafts on image failure | `preflightLocalImages` proves every local image is readable *before* anything is created (`accessSync`, not `existsSync` — an unreadable file exists but still can't be uploaded), and the missing-`--source-id` case is caught there too. If a later step fails, `discardPartialDocument` deletes the object that was created. Error messages now say "Nothing was created in Atrium" when that is true. |
| P2 | `collectImages` ignored code fences | Now runs off the same `scanLines` the validator uses. Test added with a fenced example containing both an image line and an `::atrium-asset` directive — neither is collected. |
| P2 | Dropped PDF images left dangling references | The cap is applied *before* the Markdown is written and `drop_image_references` removes the orphaned links. Verified against the real Printer & Copier PDF with the cap forced to 1: 3 extracted, 2 dropped, 1 reference remaining, and every remaining reference resolves to a file that exists. |

### CodeQL

- `js/insecure-temporary-file` (high, `run.test.js`) — **fixed**. It used a predictable `tmpdir() + Date.now()` path; now `mkdtempSync`, matching what the skill itself does.
- The two remaining medium alerts were dismissed as by-design with reasons on the alerts: `js/file-access-to-http` **is** the image upload (short-lived presigned URL from our own broker response, https-enforced, `redirect: 'error'`), and `js/http-to-file-access` **is** `get-asset`. The latter was hardened first — it now verifies the decoded bytes are genuinely PNG/JPEG/WebP before writing, so it can never write arbitrary response bytes to a caller-named path. A negative control (disabling the guard) confirmed the new test fails without it.

Zero open code-scanning alerts remain in any file this PR touches.

### Gate at `6aa3a636`

Full jest suite 4,273 passing · 103 skill tests · typecheck clean · 0 lint errors, and both touched web-tier files are lint-**warning**-free (down from 3 pre-existing warnings).

PR #1509, opened 2026-07-31
feat(ui): one design system — Meridian across the whole app, in the AI Studio palette

krishagel:
## Summary

Puts the whole authenticated application on one design system — Meridian, previously scoped to `/atrium` only — in the AI Studio palette, and fixes the UI bugs found while reviewing it.

There is no longer such a thing as an "Atrium component" and an "AI Studio component". One `.mer-*` vocabulary, one palette, one import, applied at `app/(protected)/layout.tsx` so a new page is Meridian by where it lives rather than by remembering to opt in.

## How the palette was derived

Not eyeballed. Meridian was a deep-teal system; porting it held its **contrast structure** and swapped the hue. For each token we solved for the in-gamut colour on the app's hue matching the original's ratio against its backdrop. Every relationship lands within 0.12 contrast points:

| relationship | app | original |
|---|---|---|
| ink on canvas | 15.96 | 15.84 |
| ink on surface | 17.10 | 17.02 |
| white on brand | 13.82 | 13.84 |
| ink-muted on surface | 3.13 | 3.12 |
| border on surface | 1.25 | 1.25 |
| brand on brand-tint | 12.12 | 12.13 |

**The load-bearing decision:** Whulge (`--color-primary`, `#1e6187`) is only 6.74:1 against white, but Meridian's brand plays a filled-button role assuming ~13.8:1 — every filled surface, active nav state and badge is built on a near-black brand. `--mer-brand` is Whulge's hue darkened to L=0.294. That single change is what makes the system read crisp rather than washed.

Ink stays Pacific. A deeper near-black was built, trialled side by side, and was indistinguishable in situ — so the app's own foreground won and the variant was deleted rather than left as a dead knob.

Violet remains semantic: it marks AI-agent presence and nothing else. 42 stray purple utilities were split — genuinely agent ones moved onto `--mer-agent`; the Admins shield and model-message counter went to brand, because violet was never theirs to spend.

## File layout

```
styles/meridian.css         ← import THIS
styles/meridian-tokens.css  ← the one palette
styles/meridian-core.css    ← shared structure, zero values of its own
lib/meridian/fonts.ts       ← typeface + the single portal constant
```

The barrel is not cosmetic. Core is inert without tokens, so a surface importing only core renders structure with no colour. Splitting the original stylesheet did exactly that to Atrium mid-refactor — the library rendered borderless cards on cream with no icon rail. The barrel makes importing half a design system impossible.

## Bugs fixed along the way

Several reported as "still brown" or "spacing" turned out to be something else:

- **Radix portals** mount on `document.body`, outside every scope. ~110 call sites needed the portal class; six more imported `@radix-ui/react-dialog` **directly** and rendered `<Dialog.Content>`, so a sweep for the shadcn wrapper never matched them (models edit / Add Model / Import JSON, graph node + edge forms, the assistant-architect prompt editor).
- **`SheetOverlay` and `AlertDialogOverlay` used `bg-background/80`** — a theme token on an element that portals outside any scope, so it resolved to the global cream and washed the entire viewport tan behind every sheet and destructive confirm. A scrim should be a scrim.
- **`admin/settings` saved but never closed.** Not UI: the route ran inside `withErrorHandling`, which wraps the handler's return value — but the handler returned a `NextResponse`, so the wrapper serialised it to `{"success":true,"data":{}}` and destroyed `isSuccess`, `message` and the saved row. The client checks `result.isSuccess`, got `undefined`, and took the failure branch on every successful save.
- **`Input` swallowed every layout class.** It wrapped its `<input>` in `<div className="relative">` while applying `className` to the inner element, so `flex-1`/`w-full`/`grow` were inert everywhere. The repository search field rendered ~190px wide in a 1250px card.
- **The `Card` primitive** baked `border-border/40`, a `hover:shadow` bloom and `rounded-xl` under every call site — which is why surfaces kept reading soft after their markup was cleaned. Cards are containers now; targets opt in.
- **24 cards were double-padded.** `CardContent className="pt-6"` is the upstream idiom where `Card` has no padding; here `Card` supplies `p-6` and `CardContent` adds `mt-4`, so content sat 64px from the top of its own card.
- **`/rooms/manage` "New room" and assistant-architect "Continue" did nothing** — the first reset state behind an always-visible editor, the second failed validation silently. Both are real now; the room editor and Nexus new-project are modals.

## Testing

Authenticated Playwright suite via `bun run test:e2e:local`:

**303 passed, 0 failed, 2 flaky, 40 skipped — exit 0.**

The first run was 295/6/3. All six were triaged, none left unexplained:

- Three were mine. `atrium-meridian-shell`'s leakage guard asserted `.meridian` had count 0 on `/dashboard` — the old Atrium-only invariant. It now asserts the current one: Meridian IS present off `/atrium`, Atrium's chrome is NOT, and `.meridian` is absent on public routes. `nexus-projects` and `teacher-room-create` drove create forms that are modals now.
- That last one exposed **a genuine regression this branch had shipped**: `room-feedback` renders inside the editor, so a successful save closed the modal and unmounted the banner in the same tick — the save worked and said nothing. Success feedback now renders at page level; errors stay in the modal so the form can be corrected.
- Three were environmental: expected 401, received **404** — not 429, so not the rate limiter (the harness already raises it to 600 RPM). Re-ran them in isolation against the same harness: **28 passed, 0 failed**, including the exact spec that failed 3/3 under load. A direct curl also returns 401. No API route on this branch is touched.

Also: `tsc --noEmit` clean and `eslint --max-warnings 0` clean on every touched file throughout; all 27 authenticated routes verified rendering 200, scoped, brand `#003049`, with zero nodes still painting the old cream palette.

## Risk

Four shared primitives changed — `card`, `input`, `alert-dialog`, `sheet` — which touch every surface. That is precisely why the full suite was run rather than trusted. `/(public)` and the auth routes are deliberately **not** converted and keep the global theme.


krishagel:
🤖 **pr-fix routine: blocked by a pre-existing `dev` failure — nothing to fix in this PR**

The one red check here (`Test, Lint, and Type Check`) is **not caused by this PR**. It reproduces identically on the base branch.

**Evidence**

- This PR changes **no files under `infra/`** (`git diff --stat origin/dev...HEAD -- infra/` is empty), yet all 20 type errors are in `infra/lambdas/agent-router/index.ts`.
- `dev` head (`13ae86b`) fails the same job with the **same 20 errors at the same line:col**.
- `dev` CI went red at `d0ec146` (2026-07-31 08:13 UTC, "merge: agent thread isolation and durable workspace…") and has failed on every push since. The commit immediately before it, `9e6aadb`, was green.

**Root cause**

`tests/unit/agent-workspace-identity.test.ts` (added in `b897a94`, merged as `d0ec146`) statically imports the lambda entrypoint:

```ts
import { agentRouterTestHelpers } from "../../infra/lambdas/agent-router/index"
```

Root `tsconfig.json` sets `exclude: ["node_modules", "infra"]`, but `exclude` only trims the *initial* file set — anything reachable through the import graph is still type-checked. So that one import pulls the whole lambda into the root program, where it is compiled against **root** `node_modules` instead of its own. Three consequences, all from that single cause:

| Errors | Cause |
|---|---|
| `TS2307` ×2 | `@aws-sdk/client-ecs`, `@googleapis/chat` are declared in `infra/lambdas/agent-router/package.json` only (installed at deploy/Docker build). Root `bun install --frozen-lockfile` never installs them, so they don't resolve. |
| `TS7006` ×1 | `index.ts:2572` `result.failures.map(failure => …)` — downstream of the above; `ECSClient`/`RunTaskCommand` degrade to `any`, so `failure` is implicitly `any`. |
| `TS2345` ×17 | DynamoDB DocumentClient `send()` mismatch. Root pins `@aws-sdk/lib-dynamodb ~3.1095.0` next to `@aws-sdk/client-bedrock-agent-runtime ^3.982.0`, which nest incompatible `@smithy/types` copies. The lambda's own tree doesn't have this skew. |

**Why I didn't fix it here**

The fix belongs on `dev`, not on this branch. Every open PR targeting `dev` is blocked by the same failure, so landing the repair inside a 182-file design-system PR would delay it behind this PR and mix unrelated infra surgery into a UI change. The likely minimal repair is at the test boundary — have `agent-workspace-identity.test.ts` import the helpers from a narrow module rather than the lambda entrypoint (or exclude the lambda entrypoint from the root program) — but that is live agent work landing on `dev` today and shouldn't be rewritten from a UI branch.

**No other outstanding feedback:** 0 review threads. The two bot comments are a Gemini Code Assist sunset notice and a Codex usage-limit notice — neither is actionable.

Once `dev` is green, rebase/merge `dev` into this branch and CI should clear. I've added `pr-fix-stuck` so the routine stops re-checking this PR every 3 hours; **remove the label** once `dev` is fixed and I'll pick it up again.



---

Issue #1515, opened 2026-08-01
epic: Atrium Artifact Data Service — first-party persistent data for user-built artifacts


krishagel:
## Problem

Teachers are building real classroom apps with the PSD AI Agent — a Polyatomic Ion Mahjong game, a student progress dashboard — and every one of them hits the same wall: **there is no way for a user-built artifact to persist data.**

The current workaround chain, built by a power user because nothing better exists:

1. Game writes scores to a **Google Form**
2. Form populates a **Google Sheet**
3. Agent cron reads the Sheet every 6 hours, recomputes rankings, and **bakes a JSON snapshot into the HTML**
4. Between syncs, the game falls back to **`localStorage`** so a player at least sees their own scores

Every leg of that chain is broken or fragile:

- **The 6h bake-in is stale by design.** A student who plays during 5th period sees nothing until the next cron run.
- **`localStorage` masks the failure.** In a normal tab the leaderboard "works" because it's showing you *your own device's* history. Open a private window and the leaderboard is empty — which is the honest state, and which is what the teacher reported as a bug.
- **The Sheet writes were dead anyway.** See #1514 — `sheets spreadsheets values append` was never on the broker allowlist, so the sync cron had been failing.
- **Names are free text.** The game ships a profanity filter that was substring-matching, so real surnames (Hancock, Cummings, Dickson, Sexton) were being blocked as profanity. That whole class of problem exists only because there is no authenticated identity.

### The proposed fix (Google Apps Script Web App) does not work here

A second agent proposed deploying an Apps Script Web App bound to the Sheet — "Execute as: Me", "Who has access: Anyone within psd401.net" — returning leaderboard JSON, with the game `fetch()`ing it live. That is a reasonable idea in general and it is dead on arrival in Atrium:

- Atrium artifacts render in a sandboxed iframe with `sandbox="allow-scripts"`, **no** `allow-same-origin`, and CSP **`connect-src 'none'`** (`infra/lib/atrium-sandbox-stack.ts:169`, `infra/sandbox-host/render.html`). **The artifact cannot make any network request at all.** The `fetch()` to the Apps Script URL would be blocked by CSP before it left the frame.
- The frame runs in an **opaque origin**, so `localStorage` throws there too.
- The only place the Apps Script plan would function is the raw public S3 `public-images/*.html` URL — an unauthenticated, world-readable endpoint. Putting student names and scores there is precisely the FERPA exposure everyone was trying to avoid.
[...]

PR #1541, opened 2026-08-02
feat(atrium): add sandbox AtriumData helper

### LINKED ISSUE #1518: feat(atrium): postMessage data bridge in ArtifactSandbox
Part of #1515. Depends on the server-actions sub-issue.

## Goal

Extend `components/atrium/ArtifactSandbox.tsx` so a sandboxed artifact can request data operations from its authenticated parent over `postMessage`, and get answers back.

This is the security-critical piece of the epic. The existing render-ack protocol in that file already establishes the correct pattern — follow it exactly.

## Protocol

Artifact → parent:

```js
{ type: "atrium-artifact-data-request", requestId: "<uuid>", op: "submit" | "list",
  namespace: "leaderboard", payload: {...} /* submit */, limit: 50, scope: "all" /* list */ }
```

Parent → artifact:

```js
{ type: "atrium-artifact-data-response", requestId: "<uuid>", ok: true, data: {...} }
{ type: "atrium-artifact-data-response", requestId: "<uuid>", ok: false, error: "<safe message>" }
```

## Hard requirements

- **`contentId` comes from the component's own props.** The frame never supplies it, and any `contentId`-shaped field in the incoming message is ignored, not merged. This is the difference between a data bridge and a confused-deputy hole.
- **Sender authentication is `event.source === iframeRef.current?.contentWindow`**, exactly as the existing render-ack handler at `ArtifactSandbox.tsx:210` does. The frame is opaque-origin, so `event.origin` is `"null"` and is **not** usable for authentication — do not add an origin check that would break the frame, and do not treat the absence of one as a gap (the existing comments at lines 192-210 explain why).
- **Responses go back with `targetOrigin: "*"`**, same as the existing `postMessage` at line 182, and for the same documented reason. Do not "harden" this to a concrete origin — it silently breaks delivery to an opaque-origin frame.
- **`requestId` correlation.** Responses must carry back the exact `requestId` from the request. Unknown/missing `requestId` → drop, do not respond.
- **Fail closed when there is no session.** The anonymous public reader (`/p/[slug]`) must refuse every bridge request — respond `ok: false` with a generic message, never fall through to the action. Decision 2 in #1515. Prefer a wiring-level guard (the anonymous reader does not pass the enabling prop at all) over a runtime branch, so the capability is structurally absent rather than conditionally disabled.
- **Errors returned to the frame are generic.** No stack traces, no DB errors, no distinction between "content missing" and "you can't see it" leaking through the bridge.
- Bound in-flight requests per frame so a hostile artifact cannot queue unbounded work.

## Tasks

- [ ] Extend the message handler in `ArtifactSandbox.tsx` with narrowing for the request payload (same style as the existing `isRenderAck` narrower at line 107)
- [ ] Call the server actions from the parent; map `ActionState` to the response envelope
- [ ] Thread `contentId` (and an explicit "data bridge enabled" signal) from callers of `ArtifactSandbox` — the authenticated `/c/[slug]` view enables it; the anonymous `/p/[slug]` view does not
- [ ] Update the security comment block at the top of the file to document the new op — it is the canonical explanation of this frame's trust model

## Acceptance criteria

- [ ] Jest: a message whose `event.source` is not the frame's `contentWindow` is ignored
- [ ] Jest: a request carrying its own `contentId` is served against the **prop** content id, not the supplied one
- [ ] Jest: response carries the matching `requestId`; a request with no `requestId` gets no response
- [ ] Jest: with the bridge disabled (anonymous reader wiring), every request is refused and no action is invoked
- [ ] Jest: action failure returns a generic error, and the underlying message does not appear in the response
- [ ] Existing render-ack tests still pass unchanged
- [ ] `bun run lint` zero warnings, `bun run typecheck` clean

## Files

- `components/atrium/ArtifactSandbox.tsx`
- callers of `ArtifactSandbox` (authenticated `/c/[slug]` view, anonymous `/p/[slug]` view)
- corresponding test files

### LINKED ISSUE #1519: feat(atrium): window.AtriumData helper in the sandbox host
Part of #1515. Depends on the sandbox-bridge sub-issue.

## Goal

Give artifact authors — and the agents writing artifacts for them — an ergonomic API instead of raw `postMessage` correlation code. Every artifact should call `await AtriumData.list("leaderboard")`, not hand-roll a request/response matcher.

Raw `postMessage` is the wrong surface to hand a teacher's agent: it means every generated game reimplements requestId correlation, timeout handling, and response narrowing, and each reimplementation is a new place to get it wrong.

## Implementation

In `infra/sandbox-host/render.html`, before artifact code is evaluated, define:

```js
window.AtriumData = {
  submit(namespace, payload) -> Promise<{ id, createdAt }>,
  list(namespace, options?) -> Promise<{ records: [...] }>,   // { limit?, scope?: "all" | "mine" }
}
```

- Wraps the `atrium-artifact-data-request` / `-response` protocol from the bridge sub-issue
- Generates `requestId` (`crypto.randomUUID()`), correlates responses, rejects on `ok: false`
- Per-call timeout so a dropped response rejects instead of hanging the artifact forever
- Rejects with a clear, catchable error when the bridge is disabled (anonymous reader) so an artifact can render a "sign in to see the leaderboard" state instead of appearing broken
- Listener is installed once, filters by `event.source === window.parent` and message type

## CSP — must not change

`postMessage` is not a network request. **`connect-src 'none'` stays exactly as it is** (`infra/lib/atrium-sandbox-stack.ts:169`). This is the property that makes the whole design work: artifacts get persistence without getting network access. Any PR that widens `connect-src` to make this "easier" is the wrong fix and should be rejected.

## Tasks

- [ ] Implement `window.AtriumData` in `infra/sandbox-host/render.html`
- [ ] Confirm the helper is defined before artifact code evaluates (the code arrives by `postMessage` after load — see the file's header comment)
- [ ] Add/extend the infra test asserting the sandbox CSP still contains `default-src 'none'` and `connect-src 'none'`
- [ ] `bunx cdk synth` clean; note the CloudFront/sandbox redeploy requirement on the PR (this file ships via the sandbox stack, not the app build)

## Acceptance criteria

- [ ] Infra test asserts `connect-src 'none'` is unchanged in the served CSP
- [ ] `bunx cdk synth` succeeds
- [ ] A test artifact calling `AtriumData.submit` then `AtriumData.list` round-trips through the bridge (covered by the E2E in the epic DoD)
- [ ] `AtriumData.list` rejects (does not hang) when the bridge is disabled
- [ ] `AtriumData.list` rejects on timeout when no response arrives
- [ ] `bun run lint` zero warnings, `bun run typecheck` clean

## Files

- `infra/sandbox-host/render.html`
- `infra/lib/atrium-sandbox-stack.ts` (only if the asset hash / deploy wiring needs it — do not change the CSP)
- `infra/test/atrium-sandbox-csp.test.ts` (or the existing equivalent)

## Deploy note

Requires a sandbox stack redeploy to take effect. Use the canonical dev deploy command — do not partial-deploy.

krishagel:
## Summary

Implements the sandbox-host `window.AtriumData` helper from #1519 against the locked #1518 protocol.

- Installs `AtriumData.submit(namespace, payload)` and `AtriumData.list(namespace, options?)` before artifact scripts execute.
- Uses `crypto.randomUUID()` request correlation, one exact-parent-source-filtered response listener, a 10-second per-call timeout, and a 32-call pending bound.
- Cleans pending state and timers on success, parent rejection, timeout, and synchronous post failures; rejects disabled bridge responses with a catchable `Error`.
- Captures the parent, UUID, postMessage, and timer primitives before artifact code can replace globals.
- Preserves the sandbox network boundary: `default-src 'none'` and `connect-src 'none'` are unchanged and asserted by the CDK test.

## Merge dependency

Satisfied: #1540 closed #1518 before this PR. This branch was rebased after that merge, the integrated parent/host gates were rerun, and the final head was rebased onto corrected current `dev` after #1545.

## Verification

- `bun run test:smoke:atrium-sandbox-host` — PASS, 14 checks, including graceful UUID-unavailable degradation without breaking artifact rendering and cleanup/retry after synchronous parent-post failure.
- `bun run tests/smoke/atrium-artifact-sandbox-component.smoke.ts` — PASS, 4 checks.
- `bun run test -- --runInBand tests/unit/atrium-artifact-data-bridge.test.tsx tests/unit/atrium-reader-page-masking.test.tsx tests/unit/atrium-public-reader-page.test.tsx` — PASS, 3 suites / 30 tests.
- `cd infra && bunx jest --runInBand test/atrium-sandbox-csp.test.ts` — PASS, 1 suite / 5 tests.
- `bun run lint` — PASS, zero warnings.
- `bun run typecheck` — PASS.
- `cd infra && bunx cdk synth` — PASS; existing stack-policy/configuration warnings only.
- Full authenticated Playwright E2E after #1518 — PASS, 309 passed / 40 intentionally skipped / 0 failed (349 total, 10.4 minutes).
- After the final unrelated #1545 baseline correction, the sandbox host/component/CSP suites were rerun and passed; `tests/unit/lib/repositories/content-platform-orphaned-item-sweep.test.ts` also passed (2 tests), confirming the corrected repository baseline.
- After review feedback, `bun run lint` and `bun run typecheck` were rerun on the final diff — PASS.
- Focused diff/security review — PASS, no material finding. Exact parent authentication, UUID correlation, terminal-path cleanup, pending bounds, and unchanged CSP were reviewed.

## Deployment notes

Requires a sandbox stack / CloudFront host redeploy to publish the changed `render.html`. Use the canonical full dev deployment with `AIStudio-AtriumSandboxStack-Dev` included; do not partial-deploy the sandbox stack. No deployment, production setting, or production data operation was performed by this PR.

Closes #1519

Part of #1515

### DIFF
--- a/infra/sandbox-host/render.html
+++ b/infra/sandbox-host/render.html
@@ -29,6 +29,10 @@
   script-src 'unsafe-inline'). We render only ONE message (the first valid one
   is replaced on each subsequent valid message) and never eval arbitrary strings
   ourselves beyond reconstructing the author's own script nodes.
+
+  Before any artifact code runs, this host installs window.AtriumData. The helper
+  uses postMessage to ask the authenticated parent for scoped data operations;
+  it does not weaken the frame CSP or give artifact code direct network access.
 -->
 <html lang="en">
   <head>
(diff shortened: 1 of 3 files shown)


krishagel:
Addressed the UUID-availability review finding in `620c1d7b0`: the host now feature-detects `crypto.randomUUID`, keeps `window.AtriumData` installed, rejects calls with a clear unavailable error when UUID generation cannot initialize, and still registers/runs artifact rendering. Added a regression smoke for that degraded runtime; the host smoke now passes 13 checks, and final full lint/typecheck both pass.

krishagel:
Final review disposition in `adece98f3`: corrected the response-listener comment to describe exact WindowProxy pinning accurately, and added a synchronous parent-`postMessage` failure/retry smoke that verifies cleanup. The host smoke now passes 14 checks; full lint/typecheck pass. The `"*"` target remains unchanged because it is the locked #1518 opaque-sandbox protocol and the receiver authenticates by WindowProxy identity.

Issue #1517, opened 2026-08-01
feat(atrium): server actions for artifact record submit/list


krishagel:
Part of #1515. Depends on the schema sub-issue.

## Goal

Server actions that are the **only** way records reach `content_data_records`. All authorization, identity, and bounding lives here — the bridge in `ArtifactSandbox.tsx` is a dumb transport and must not carry any trust.

## Actions

Both in `actions/db/atrium/artifact-data.ts` (matching the existing per-file action layout under `actions/db/atrium/`).

### `submitArtifactRecord({ contentId, namespace, payload })`

- Session auth via `getServerSession()`; no session → `authNoSession`
- Resolve the content through the existing `contentService.get` → `visibilityService.canView` path (same as `actions/db/atrium/get-content.ts`) so a caller who cannot view the content gets a **404-style NotFound, not a 403** — visibility check before permission check, per the 404-masks-existence guard order
- `user_id` is taken from the session. **Never** accept a user/name/identity field in `payload` position — reject the call if the payload contains a reserved key (`userId`, `user_id`)
- Validate: `namespace` matches `^[a-z0-9_-]{1,64}$`; serialized `payload` ≤ 8KB
- Rate limit per user using the existing `lib/rate-limit.ts`. Per decision 3 in #1515 this is a sanity guard against a runaway `requestAnimationFrame` loop, not a quota system — pick a generous limit and do not build metering
- Returns `ActionState<{ id, createdAt }>`

### `listArtifactRecords({ contentId, namespace, limit?, scope? })`

- Same session + visibility gate as above
- `limit` bounded to ≤ 200, default something small (50)
- `scope`: `"all"` (default — every user's records for this content+namespace, which is what a leaderboard needs) or `"mine"` (just the caller's)
- Returns records ordered `created_at DESC`, each with the submitting user's **display name** resolved server-side and the record payload. Per decision 1 in #1515, real session names — no aliasing layer
- Returns `ActionState<{ records: [...] }>`

## Non-goals

- No update or delete op in this issue. Append + list only. If an artifact needs "replace my high score," it appends and the read path takes the max — that is simpler and auditable.
- No anonymous path. Decision 2 in #1515.

## Acceptance criteria

- [ ] Both actions follow the server-action template (requestId, timer, `createLogger`, `handleError`, `createSuccess`)
- [ ] Unit test: submit succeeds for a user who can view the content
[...]

PR #1542, opened 2026-08-02
feat(agent): add Atrium artifact data reads

### LINKED ISSUE #1520: feat(agent): Atrium broker read path for artifact data records
Part of #1515. Depends on the schema and server-actions sub-issues.

## Goal

Let the PSD AI Agent read `content_data_records` through the Atrium broker, so a teacher can ask their agent to build a progress dashboard from the same records the game writes — instead of scraping a Google Sheet.

This is what closes the loop on the original use case. The teacher's student-dashboard proof-of-concept had to ship with **synthetic data** because the agent had no readable data source. With this, the dashboard reads real records.

## Implementation

Add to `ALLOWED_PATHS.GET` in `app/api/agent/atrium/route.ts`:

```
new RegExp(`^/${IDENTIFIER}/data$`)
```

Handler behavior:

- Query params: `namespace` (required, same `^[a-z0-9_-]{1,64}$` validation), `limit` (bounded, ≤ 200), optional cursor/offset for paging
- Authorization goes through the existing owner-operation path (`lib/agent-workspace/atrium-owner-operation.ts`) — the agent reads as its **owner**, so the same content visibility gate applies. An agent must not be able to read records for content its owner cannot view.
- Read-only. **No POST/PATCH/DELETE for data records in this issue** — agents write artifacts, students write records. Keeping the write path exclusive to the sandbox bridge means every record has a real session user behind it.
- Response shape mirrors `listArtifactRecords` so the same consumers work.

## Acceptance criteria

- [ ] `GET /<id>/data?namespace=leaderboard` returns bounded records for an owner who can view the content
- [ ] Same request returns the 404-style response for an owner who cannot view the content (visibility before permission)
- [ ] Missing or malformed `namespace` → 400 with a structured error naming the constraint
- [ ] `limit` above the cap is clamped, not honored
- [ ] `POST`/`PATCH`/`DELETE` to `/<id>/data` remain rejected by the path allowlist — explicitly tested
- [ ] Route security tests follow the existing pattern in the agent atrium route test file
- [ ] `bun run lint` zero warnings, `bun run typecheck` clean, jest green

## Files

- `app/api/agent/atrium/route.ts`
- `lib/agent-workspace/atrium-owner-operation.ts`
- corresponding test files

krishagel:
## Summary

- allow only `GET /<id>/data` through the signed-owner Atrium broker
- validate the required `namespace`, default reads to 50 records, and clamp caller limits to 200
- preserve the shared `contentService.get` 404 existence mask before any record query or authoring-permission gate
- return the #1517-compatible `records[{ id, displayName, payload, createdAt }]` shape without user IDs or emails
- keep `POST`, `PATCH`, and `DELETE /<id>/data` outside the fixed allowlist
- add focused authorization, malformed-input, limit, existence-masking, content-kind, response-shape, and method-rejection coverage

## Verification

- `bun run lint` — passed with zero warnings
- `bun run typecheck` — passed
- `bun run test:ci` — 563 suites passed, 1 skipped; 5,272 tests passed, 15 skipped (5,287 total)
- `bun run test:ci --runInBand tests/unit/agent-atrium-route.test.ts tests/unit/agent-atrium-owner-operation.test.ts` — 2 suites / 59 tests passed on the review-follow-up head
- focused diff/security review — no material authorization, masking, bounding, privacy, or method-surface findings remained

The broad local pre-push E2E hook was bypassed only because this managed worktree has no `AUTH_SECRET`. Issue #1520 requires Jest, lint, and typecheck; it does not require an E2E or CDK command.

## Deployment notes

No schema migration, infrastructure change, deployment setting, API v1 documentation, or agent skill documentation change is included.

Closes #1520

Part of #1515

### DIFF
--- a/app/api/agent/atrium/route.ts
+++ b/app/api/agent/atrium/route.ts
@@ -12,6 +12,8 @@ const ALLOWED_QUERY_KEYS = new Set([
   "tag",
   "query",
   "since",
+  "namespace",
+  "limit",
 ])
 const IDENTIFIER = "[A-Za-z0-9%._~-]{1,384}"
 
@@ -30,6 +32,7 @@ const ALLOWED_PATHS: Record<AtriumBody["method"], readonly RegExp[]> = {
     // Committed markdown source — the ONLY way an agent can read a document's
     // body text (`GET /<id>` returns bodyLocation "proof" with no text).
     new RegExp(`^/${IDENTIFIER}/source$`),
+    new RegExp(`^/${IDENTIFIER}/data$`),
     // Authored image assets (#1284): metadata list, plus a bounded
     // base64 byte read so an agent can copy an image between objects.
     new RegExp(`^/${IDENTIFIER}/assets$`),
--- a/lib/agent-workspace/atrium-owner-operation.ts
+++ b/lib/agent-workspace/atrium-owner-operation.ts
@@ -1,3 +1,4 @@
+import { and, desc, eq } from "drizzle-orm"
 import { z } from "zod"
 import {
   ApprovalRequiredError,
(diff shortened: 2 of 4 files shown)


krishagel:
Addressed the actionable coverage gap in da4d07e8e: visible non-artifact content now has an explicit test proving structured 400 behavior before record access or the authoring gate.\n\nThe remaining observations are intentional scope decisions: #1520 specifies namespace + bounded limit (not scope), and #1517 defaults to all records; broker-wide rate limiting, payload semantics, and a generalized shared repository helper are outside this locked small issue. No production change was made for those non-regressions.

krishagel:
Confirmed: the lack of a new per-route rate limiter is intentional for #1520. The locked acceptance criteria define the broker bound as the record limit (default 50, clamp at 200) and do not introduce a new rate policy. Existing broker reads (list/get/source/assets) share the same invocation boundary without per-operation rate limiting. Adding a one-off policy only for data reads would be a separate broker-wide design decision, so this issue keeps the established surface consistent.

PR #1606, opened 2026-08-05
fix(chat-chart): restore the local render engine so student data can be charted (#1596)

### LINKED ISSUE #1596: Visualize achievement/assessment data without exposing student data to third-party services
Context: while helping a user visualize Smarter Balanced achievement-gap data by race/ethnicity at a school via the psd-data warehouse skill, the chat-chart skill's --sensitive flag hard-refused because the local on-host chart engine is disabled in this build, with error: chat-chart --sensitive is set and the local on-host engine is disabled in this build, so the only available engine QuickChart would transmit the data to third-party quickchart.io, refusing to render sensitive data off-district, restore the local engine to chart sensitive data, exit code 3. This means there is currently no supported way to render an inline chart image for sensitive or student data in Google Chat. The only fallback is a hand-built HTML canvas page via the canvas skill, which only renders on the web or app Control UI surface, not in Google Chat, so Chat users see raw embed directive text instead of a chart. Impact: achievement-gap, attendance, grade-distribution, and similar equity or data-analysis requests are common and high-value for district staff, but the current tooling forces a choice between sending student-derived aggregate data to a third-party service via the only working sensitive path, or producing no chart at all in the primary chat surface. Requested fix: restore or enable the local matplotlib-based chart engine for chat-chart, the skill already supports this via engine local flag, it is just disabled in this build, so sensitive data can be charted without leaving PSD AWS environment. Alternatively provide a Chat-native way to render canvas or HTML visualizations so achievement-data visualizations work in the primary surface. Reported by cantonwinej at psd401.net via agent request, 2026-08-04.

krishagel:
## Description

Closes #1596.

A staff member asked for a Smarter Balanced achievement-gap chart by race/ethnicity in Google Chat and got this instead of a chart:

```
chat-chart --sensitive is set and the local on-host engine is disabled in this
build, so the only available engine QuickChart would transmit the data to
third-party quickchart.io, refusing to render sensitive data off-district
exit code 3
```

The refusal was correct — QuickChart puts the values in a third party's logs — but it left no supported way to chart student data in the district's primary chat surface. The local engine had been pulled after the 2026-05-18 AgentCore incident (matplotlib in the venv + an npm install for the S3 upload; the microVM overlay mount failed and every boot died).

**This restores the local engine as a pure-Node rasteriser with zero image weight**: no pip dependency, no npm dependency, no new Dockerfile `RUN`, no new layer. It ships via the existing `COPY skills`, so it cannot reintroduce either historical failure mode — there is no Python file to leave a `__pycache__` beside (the 2026-05-18 root cause, per `.dockerignore`) and no install step to spend a layer on (the 2026-07-02 root cause).

### What changed

- **`render_local.js` (new)** — bar/line/pie/scatter drawn into an RGB buffer at 2× and box-downsampled to an 800×500 PNG, encoded with `node:zlib` plus a hand-rolled CRC32 (`zlib.crc32` needs Node ≥ 22.2 and the agent image's Node comes from an unpinned base image — a missing built-in would break every chart at render time). Labels use a built-in 5×7 ASCII font. ~100 ms per chart.
- **`render_local.py` deleted**, along with the skill's vestigial `@aws-sdk/client-s3` dependency — uploads already go through the agent broker over plain `fetch`.
- **`run.js`** renders in-process (no `python3` subprocess, no temp file, no `validated-fs`). `--user` is now optional: the broker derives the storage path from the caller's identity, so the email was provenance only.
- **Docs** — SKILL.md engine table/limits/failure modes, the ops runbook rows, and the Dockerfile/`.dockerignore` comments that told the next reader to re-add matplotlib.

### Engine policy, with REV-INFRA-002 intact

| Engine | Behaviour |
|---|---|
| `auto` (default), `local` | Renders on-host. Values never leave PSD AWS, so sensitivity no longer decides whether a chart exists at all. |
| `quickchart` | Runs **only** when named explicitly **and** the data is neither `--sensitive` nor PII-matching. Otherwise exit 3 — never a silent downgrade, never a URL built. |

`--sensitive` still matters (it documents intent and hard-blocks an accidental `--engine quickchart`), it just is no longer the difference between a chart and a refusal.

## Verification

**End-to-end, the issue's exact scenario** — `--sensitive` achievement-gap data, driven through the real CLI against a stub broker. Previously exit 3 with no chart; now exit 0, a rendered PNG uploaded through the broker, and a `PSD_AGENT_RICH_V1` card envelope on stdout. Any reviewer can reproduce every image in ~10 seconds with no dependencies:

```bash
cd infra/agent-image/skills/chat-chart
node -e "const {renderChartPng}=require('./render_local.js');const {buildChartJsConfig}=require('./run.js');
require('node:fs').writeFileSync('/tmp/chart.png', renderChartPng(buildChartJsConfig('bar',
  [{label:'Am. Indian',value:38.2},{label:'Asian',value:71.4},{label:'Black',value:44.9},
   {label:'Hispanic',value:52.1},{label:'2+ Races',value:58.7},{label:'White',value:66.3}],
  'SBA ELA proficiency by race/ethnicity')))"
```

Rendered and inspected during this change: the achievement-gap bar chart above, a 5-day attendance line (0.938–0.962 — the axis keeps the data's own range so the trend is visible rather than flattened against zero), a 4-slice level-distribution pie with legend, a scatter, a 50-point stress chart, a 14-date line, a 25-slice pie, an 80-character title, and negative values beside 2.45M magnitudes.

**Tests: 44, in `run.test.js`** (all new tests live in that file deliberately — CI's test-file list is in `.github/workflows/ci.yml`, which this routine must not edit). `bun test infra/agent-image/skills/chat-chart/run.test.js` → 44 pass, and `node --test` agrees.

They are **mutation-tested**. The starting assertions (`inked > 2000`) were satisfied by the title and axis furniture alone — "pie wedges never painted", "scatter axes swapped", "line segments dropped" and 7 other mutations all left the suite green. The current suite catches every one, plus every fix below reverted individually.

### Defects found and fixed during self-review

Security:
- **`niceTicks` could never terminate** — with a step below an ULP of the axis start, `value += step` is a no-op and the tick array grew until V8 aborted on the heap limit. Reachable on the **default** engine with no flags, from inputs as ordinary as two nanosecond-epoch timestamps 300 ns apart. The loop is now indexed and the degenerate range is refused.
- **The PII gate scanned raw argv, not the payload QuickChart receives** — `"jsmith@psd401.net"` sailed past the regex until `JSON.parse` restored it, and `--title` was never scanned despite appearing in the QuickChart URL twice. The gate now reads the built config.
- **Quadratic backtracking in the email detector** — 5.2 s of CPU to reject a 60 KB label, ~25 s at the 120 KB argv allows. Bounded quantifiers: 0.10 s.
- **"Data never leaves PSD AWS" overstated the guarantee** — the values don't, but the rendered PNG is published to an unauthenticated (unguessable, ~30-day) workspace URL because Google Chat has to fetch it. Now stated plainly in SKILL.md and the runbook rather than implied away.

Correctness:
- Titles of 45+ characters were clipped at **both** ends (52-char budget overflowed the canvas; the text is centred). Budget is now derived from the canvas width.
- `formatNumber` kept one decimal in the k/M branches — coarser than the tick steps those axes use — so a 1.00M–1.10M axis printed "1.0M" six times, and bars of 1.02M and 1.04M were labelled identically. Also added B/T suffixes and an exponent fallback so 1e12+ no longer runs off the left margin.
- Tick values rounded with `toFixed(10)` collapsed every tick below ~1e-6 onto one number, leaving an axis that didn't cover its own data.
- Pie legend rows ran off the canvas past 18 slices, silently taking those slices' identities with them (a wedge has no other label). Rows now shrink to fit and, when they can't, the last row says how many are unlabelled.
- Dense category axes printed one-character stubs; they now keep whole labels and letter every Nth category. `Infinity` passed validation and serialised into the QuickChart URL as `null`. `--user` was validated only on the local path. `truncateLabel` returned 2 characters for a 1-character budget.

## Checklist

- [x] No `console.*` calls in production/shared code — the skill writes to `process.stdout`/`process.stderr` by design (it is a CLI, outside the Next.js runtime and outside `@/lib/logger`'s reach)
- [x] **No imports or usage of `@/lib/logger` in client components or client hooks** — not applicable, no client code touched
- [x] Client code uses `console.error` only for actionable errors — not applicable
- [ ] Lint passes (`bun run lint`) — **could not run here**: `bun install` fails with HTTP 403 from npm through this routine's egress policy, so ESLint isn't installable in the sandbox. Checked by inspection against `eslint.config.mjs` instead, including the rules CI enforces at `--max-warnings 0`: `max-params` (the drawing primitives were refactored to rect/point/position objects for this), `complexity`, `max-depth`, `max-lines-per-function`, `no-console`, and the enabled `unicorn` subset (`prefer-node-protocol`, `number-literal-case`, `prefer-number-properties`, `explicit-length-check`, `no-for-loop`, `throw-new-error`). CI is the real gate here.
- [x] All tests pass — chat-chart 44/44 under both Bun and Node; `_shared/agent-broker`, `psd-data`, `psd-aistudio`, `psd-brand-guidelines`, `psd-skills-meta`, `psd-summarize` and the 69-test `test_eval_runner.py` suite all still pass. (`infra/lambdas/nexus-conversation-retention` fails to load `@aws-sdk/client-secrets-manager` — same blocked-registry cause, pre-existing and unrelated.)
- [x] Types/interfaces follow project conventions — CommonJS agent skill, matching its siblings
- [x] No type assertions — no TypeScript touched
- [x] Database field names — not applicable
- [ ] TypeScript strict mode errors resolved — **could not run** `bun run typecheck` (same blocked registry); no TypeScript files are changed by this PR
- [x] Environment variables handled securely — none added
- [x] Documentation updated — SKILL.md, `docs/operations/agent-platform-setup.md`, Dockerfile and `.dockerignore` comments
- [x] App builds and runs without module errors — the skill has no `node_modules` at all now; it loads and runs on Node 22 and Bun with an empty dependency tree

### Known gap, stated rather than hidden

The only chat-chart eval (`evals/synthetic-bar-chart.yaml`) exercises the QuickChart path, so the engine that now serves ~100% of real traffic has no eval coverage. Adding one needs `eval/broker_stub.py` to implement the `workspace-storage` publish/complete-upload contract, which it doesn't today — that's new eval-harness capability I can't validate without a live agent, and landing an unvalidated entry in the regression suite would mean a false red. The unit suite covers the renderer at pixel level instead.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PQQwWxSorXQoCz5wnHmkMD)_


krishagel:
🤖 pr-fix routine round 1 complete. Pushed `66742402` (2 commits).

**CI failures fixed: 2**

| Check | Cause | Fix |
|---|---|---|
| Test, Lint, and Type Check | `security/detect-unsafe-regex` at `run.js:52` | Flattened the email PII pattern |
| CodeQL (1 new error alert) | Same line — the nested group's `[A-Za-z0-9-]` overlaps the `[A-Za-z]` TLD after it, so labels can be redistributed across the boundary (polynomial backtracking) | Same fix |
| Test, Lint, and Type Check | `no-redeclare` at `run.test.js:324` | See below |

Bounding the email quantifiers (the earlier ReDoS fix) needed a nested `(?:\.[A-Za-z0-9-]{1,63}){0,4}` to keep matching multi-label domains — star height 2, which safe-regex rejects outright. It is now one flat trailing class with nothing after it to backtrack against: star height 1, linear. Measured on a 120KB payload: **24 ms** with and without an `@`, versus ~5 s for the original unbounded pattern on *half* that input. It accepts a marginally wider shape (1-char TLD, trailing dot), which is the safe direction for a gate whose false positives cost an on-host render — the default anyway.

**Review comments addressed: 4 of 5**

- **`no-redeclare` / duplicate `countNonWhite`** — the second declaration hoisted over the first, so the first was already dead and every call site resolved to the survivor. Deleted the dead one; no behaviour change.
- **Duplicate test name** — kept the version asserting ink in `X_LABEL_BAND`, deleted the other. Its `countNonWhite(...) > 2000` is exactly the assertion your mutation note showed is satisfied by title and axis furniture alone, so nothing was lost.
- **CRC32 test oracle** — good catch, and the failure mode was the sharp one: on a runtime without `zlib.crc32` the test threw, reporting its own assumption as a renderer defect. Replaced with a bit-at-a-time reference in the test file — no lookup table, so a mistake building production's table cannot cancel out against it — pinned to the CRC-32/ISO-HDLC check values so a broken oracle fails loudly instead of blessing a broken renderer.
- **No CLI-level test of the local path** — the real gap, agreed. Added three tests against a stub broker on the port `agent-broker.js` hardcodes: the default path (args → `renderChartPng` → `publishArtifact` → stdout envelope), issue #1596's exact `--sensitive` achievement-gap scenario (previously exit 3, now a rendered PNG), and a broker failure exiting non-zero rather than printing a card. They assert the uploaded bytes are a real PNG and that the checksum the broker signed actually covers them. Note for future readers: these use async `spawn`, not `spawnSync` — `spawnSync` blocks the event loop, so an in-process stub server never gets to answer the child.
- **SKILL.md "50 points is a hard ceiling"** — scoped to the engines that enforce it, with a note that `--engine quickchart` has no point-count check and is bounded only by argv and the ~16KB spec-URL ceiling.

**Not taken:** the binary search for `wedgeColour`'s per-pixel `bounds.find()`. With the 50-point cap and a ~100 ms budget it is not on the critical path; left a comment-level note for if a many-slice pie ever shows up slow.

**Verification:** 47 pass / 0 fail under both `node --test` and `bun test`. Mutation-tested rather than assumed — corrupting the CRC polynomial fails the chunk test, and changing the upload content type fails the CLI test.

Lint could not be run here (`bun install` still 403s against npm through this routine's egress), so the two lint fixes are checked by inspection against `eslint.config.mjs` — including `max-lines-per-function` (150; largest addition is 47), `max-params` (5; all additions ≤ 1), `max-depth`, `complexity`, and `unicorn/{prefer-node-protocol,number-literal-case,no-for-loop}`. CI is the real gate; I'll pick this up again next fire if it disagrees.



---

krishagel:
Superseded by #1607, which merges this branch in and fixes the two CI blockers that this environment could not reproduce (`bun install` returned HTTP 403 through its egress policy, so lint and typecheck never ran here):

- `security/detect-unsafe-regex` on the email PII backstop — bounding the quantifiers was not sufficient. The trailing component stayed ambiguous (the label class is a superset of the TLD class and both follow a dot), and a quantifier nested inside a quantified group is star-height 2 whatever the bounds. Flattened to one level; detection unchanged, 60KB adversarial input rejects in 27ms. This was almost certainly the CodeQL alert too (polynomial ReDoS on the same pattern).
- `no-redeclare` on `countNonWhite` — two declarations shared the name; hoisting meant the second was already the one every call resolved to, so the first was dead.

The render engine work here is unchanged and lands as-is. #1607 also carries fixes for nine other root causes from the same prod failure analysis. Closing in favour of that PR.

PR #1608, opened 2026-08-06
fix(agent): name failed-turn side effects, allow access-proposal resolution

### LINKED ISSUE #1596: Visualize achievement/assessment data without exposing student data to third-party services

krishagel:
## Description

The last two fixable items from the 2026-08-03..05 prod failure analysis (#1607). Both were things I previously handed back as "needs you" — neither actually needed you.

**1 · A failed turn now names the actions that already ran** — 11 rows / 7 users

`OpenClawIncompleteToolTurn` told the user only *"some tool actions may have already been executed — please verify before retrying."* A principal reading that can't tell whether a doc was created, a calendar event booked, or nothing at all happened, so the only safe response is to check everything by hand.

An automatic retry stays out, and that part is a real constraint rather than caution: by the time this class is raised the OpenClaw host has already attempted its one tools-disabled finalization, so replaying could repeat effects that landed. But the harness records every completed tool call, so it can name them — and now does, across this and the other failed-partial paths.

The most valuable case is the empty one: **when nothing completed, the message says so and tells the user it's safe to retry**, instead of sending them hunting for side effects that don't exist. Names are deduplicated and capped at five so a turn that looped over forty files doesn't paste forty entries into Chat; an unrecognized tool-call shape falls back to today's conservative wording rather than claiming nothing ran.

**2 · `drive accessproposals resolve` is allowlisted** — agent_failures 1986

While Drive sharing was broken, a user submitted a Drive access request for an agent-owned doc and the agent tried to accept it — the one recovery path available from the other direction. That operation wasn't in the broker allowlist at all, so it was refused too, and the user was left with no route to the file.

Resolving a proposal grants the requester access, so it's held to the same ceiling as a named-person share: **reader, commenter or writer, never owner**. `deny` needs no role. Drive accepts the role as a bare string or a single-element list; both are handled. Agent slot only, allowlisted keys, fails closed on an unparseable payload.

## Not fixable in this repo

**`students_demographics` timeouts** (opeln — 5 attempts, 15–100s, including a bare `SELECT COUNT(*)`). `psd-data` queries execute against an external `psd-data-mcp` server backed by Redshift. That table, its row-level-security predicate, and any sort/dist-key tuning live outside this repository — there is no schema here to change. A bare `COUNT(*)` timing out points squarely at the RLS predicate on that table, which is a warehouse-side fix.

The shared-calendar item (olearyd) is a Google sharing change — the agent account needs write access to the HRMS Staff Calendar. Not code, and confirmed as user-side.

## Verification

- Full unit suite: **571 suites / 5,374 tests passing**
- Complete agent-image Python suite exactly as CI invokes it (all 20 files, `unittest`, Python 3.12): **702 tests, OK**
- **E2E: 319 passed** via the pre-push gate

New coverage: six cases on the failed-turn message (names completed tools, declares a safe retry when none ran, ignores in-flight calls, dedupes and caps, preserves the partial answer, keeps conservative wording on an unknown shape) and eleven on access proposals (each allowed role, list-form role, denial without a role, plus owner/organizer/no-role/unknown-action/unknown-key/unparseable/user-slot refusals).

## Housekeeping

Closed **#1596**, which #1607 left dangling: #1606 carried the `Closes #1596` keyword but was closed in favour of #1607 rather than merged, so GitHub never fired the auto-close, and #1607's description didn't repeat it. Verified the work is on `dev` (`render_local.js` present, `render_local.py` removed) before closing.

## Checklist

- [x] No `console.*` in production code — agent skills are CLIs outside the Next.js runtime
- [x] Typecheck passes
- [x] All tests pass — unit, agent-image Python, and E2E
- [x] No `any` types; no new type assertions

### DIFF
--- a/infra/agent-image/harness_adapter.py
+++ b/infra/agent-image/harness_adapter.py
@@ -246,6 +246,7 @@ def _classify_chat_error(error_message: str) -> str:
 def _frame_failed_partial(
     partial: str,
     completed_tools: Optional[List[Dict[str, Any]]] = None,
+    in_flight_tools: Optional[Dict[str, Any]] = None,
 ) -> str:
     """Wrap a failed/degraded turn so it is never presented as a clean answer.
 
@@ -264,7 +265,7 @@ def _frame_failed_partial(
     nothing at all happened. We KNOW which tools completed, so say so.
     """
     partial = (partial or "").strip()
-    ran = _describe_completed_tools(completed_tools)
+    ran = _describe_completed_tools(completed_tools, in_flight_tools)
     if partial:
         return (
             "⚠️ I couldn't finish that — I hit a problem partway "
@@ -278,6 +279,7 @@ def _frame_failed_partial(
 
 def _describe_completed_tools(
     tool_calls: Optional[List[Dict[str, Any]]],
+    in_flight: Optional[Dict[str, Any]] = None,
 ) -> str:
     """Render the finished tool calls as a short clause, or '' when none ran.
 
@@ -297,6 +299,25 @@ def _describe_completed_tools(
     # created — precisely the failure this function exists to prevent. So an
     # unreadable terminal record suppresses the safe-to-retry claim.
     unnamed_terminal = False
+    # A tool that STARTED and never reported back is the most dangerous case,
+    # not the safest: the request may have reached the broker and created the
+    # Doc before the turn died, and its result event simply never arrived.
+    # Started calls live in `tool_starts` and are removed from it when they
+    # complete, so anything still there is genuinely in flight — and it never
+    # appears in `tool_calls` at all, which is how an interrupted side effect
+    # could leave that list empty and produce a "safe to retry".
+    started_names = []
+    if isinstance(in_flight, dict):
+        for start in in_flight.values():
+            if not isinstance(start, dict):
+                unnamed_terminal = True
+                continue
+            name = start.get("name")
+            if isinstance(name, str) and name and name != "unknown":
+                if name not in started_names:
+                    started_names.append(name)
+            else:
+                unnamed_terminal = True
     for call in tool_calls:
         if not isinstance(call, dict):
             unnamed_terminal = True
@@ -309,18 +330,33 @@ def _describe_completed_tools(
                 names.append(name)
         else:
             unnamed_terminal = True
+    # An in-flight call is reported as uncertain rather than as completed: we
+    # know it was requested, not whether it landed.
+    if started_names:
+        started_shown = ", ".join(started_names[:3])
+        started_more = (
+            f" (+{len(started_names) - 3} more)" if len(started_names) > 3 else ""
+        )
+        started_note = (
+            f" {started_shown}{started_more} was still running and may or may not "
+            "have completed."
+        )
+    else:
+        started_note = ""
+
     if not names:
-        if unnamed_terminal:
+        if started_names or unnamed_terminal:
             return (
-                " Some steps may have already run, so please check before retrying."
-            )
+                f"{started_note} Some steps may have already run, so please check "
+                "before retrying."
+            ).lstrip()
         return " Nothing had run yet, so it's safe to retry."
     shown = ", ".join(names[:5])
     more = f" (+{len(names) - 5} more)" if len(names) > 5 else ""
     unknown_note = " (and possibly others)" if unnamed_terminal else ""
     return (
         f" I had already run: {shown}{more}{unknown_note} — "
-        "please check those before retrying."
+        f"please check those before retrying.{started_note}"
     )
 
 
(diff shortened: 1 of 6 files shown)


krishagel:
All six findings addressed in `d98b95b79` — every one was real.

**@claude's security finding was the important one.** `validateAccessProposalResolve` only validated `role[0]`, so `{"role":["reader","owner"]}` passed on its first element — and as you noted, `executeWorkspaceCommand` forwards the original argv to `gws` verbatim, so nothing downstream re-reads the array. The whole list is now validated and must contain exactly one element. Covered by three new refusal cases (multi-element, empty, non-string element).

**The status-normalization finding was the same bug this PR exists to fix, reintroduced.** You were right that the legacy `tool_result` path doesn't normalize `status`, so a real completion arriving as `"completed"` would be dropped and the message would claim a safe retry after a Doc had already been created. Fixed from both ends: status is normalized at the legacy ingestion site the way the item path does it, and the filter now excludes only genuinely *pending* statuses — an unknown terminal status counts as completed, because assuming otherwise is the unsafe direction. @codex's related P1 is the same root: an unreadable terminal record (no name, or `"unknown"`) now suppresses the safe-to-retry claim outright, and alongside named tools the message says "and possibly others".

**@codex's test-ordering P2 was a genuine catch** — the new class sat after `unittest.main()`, so direct execution ran main() before the class existed and silently skipped all six cases. They only ran because CI invokes `-m unittest <file>`. Moved above the `__main__` guard; direct execution now reports 81 tests.

Also fixed: the contradictory "safe to retry" + "please check before retrying" wording, the undocumented operation (@claude was right that allowlisting it without a SKILL.md example left it functionally inert — the model only emits what the skill tells it about), skill-layer coverage beyond the broker validator, and both minor notes (type annotations, the four-quote docstring).

One note on E2E: `nexus-projects.functional.spec.ts` failed on the first pre-push run and passed on rerun in 3.4s against a 2-minute timeout. My diff is six files under `infra/agent-image` and `lib/agent-workspace` — nothing near Nexus project membership. Treating it as flake, with the full suite green at 319 passed.

Verification: full unit suite 571 suites / 5377 tests, complete agent-image Python suite as CI invokes it (705 tests), direct execution of test_harness_adapter.py (81), psd-workspace 83, command-executor 133, E2E 319 passed, lint clean at `--max-warnings 0`, typecheck clean, bootstrap budget passing.

krishagel:
@codex review

Head is `d98b95b790`. Both of your P1s and both P2s are addressed there, along with @claude's role-array bypass — details in the comment above.

The two worth re-checking closely:

- **`_describe_completed_tools` / `_normalize_tool_status`** (`harness_adapter.py`) — the filter now excludes only genuinely *pending* statuses rather than allowlisting `success`/`error`, so an unknown terminal status counts as completed. An unreadable terminal record (missing or `"unknown"` name, or a non-dict entry) suppresses the safe-to-retry claim entirely rather than being dropped. Worth confirming there is no remaining path where a real side effect yields "safe to retry".
- **`validateAccessProposalResolve`** (`command-executor.ts`) — the role list is validated in full and must have exactly one element, closing the `["reader","owner"]` bypass.

krishagel:
@codex review

Head is `44dc63d3c`. Both P1s fixed.

**In-flight tools are now retry-unsafe.** You were right, and this was the third instance of the same unsafe direction in this PR — the previous commit's own test asserted the wrong behaviour ("ignores tools still in flight"), locking the bug in rather than catching it. `tool_starts` is now passed alongside `tool_calls` at all three call sites; started calls are named and reported as "was still running and may or may not have completed", an unnamed in-flight call blocks the claim too, and safe-to-retry now requires **both** lists empty.

**The documented argv could not have worked.** `fileId`/`proposalId` were in `--json` when they address the endpoint and belong in `--params` — the model would have been taught an invocation that fails. Split correctly in SKILL.md, and the validator now judges the union of `--params` and `--json` so an unrecognized key is caught wherever it sits and the role ceiling holds regardless of which flag carries it. Tested against the exact documented argv.

PR #1612, opened 2026-08-06
fix(nexus-memory): stop refusing the user's own personal information, and stop losing whole imports to a truncated model response

### LINKED ISSUE #1610: Nexus memory import: PII gate silently rejects candidates (no reason shown); extraction returns invalid response on large pastes
## Summary

On 2026-08-05 between ~11:22am and ~12:40pm PDT, three prod users tried the new Nexus memory import (Epic #1406 / #1409) and it was spotty to fully broken. Prod CloudWatch logs (`/ecs/aistudio-prod`) show two independent defects: (1) the fail-closed PII gate in `memoryService.save` deterministically rejects imported candidates that contain any name/date/age/contact detail — which vendor memory exports are naturally full of — and the UI never tells the user why, so one user retried the same doomed batches 8+ times over an hour; (2) the extraction model call became unparseable for the last 3 attempts of the day, so extraction returned a hard error and no candidates at all.

## What the prod logs show (all times PDT, 2026-08-05)

**Extraction phase** (`extractImportCandidates`) — 8 attempts all day, 5 success / 3 error:

| Time | Result | Duration | Detail |
|------|--------|----------|--------|
| 11:22–11:53 | 5 successes | 3.3–5.7s | users 77, 224, 633; vendors chatgpt/claude; 13–28 candidates each |
| 11:55, 11:57, 12:02 | 3 errors | 27.5s, 18.0s, 24.7s | `The memory extraction model returned an invalid response` (requestIds `vx9yCtz79E`, `zYH2i1Yxvp`, `SM-OgRKrfR`) |

No successful extraction occurred after 11:53am — the last three attempts of the day all failed.

**Save phase** (`saveImportedMemories`) — every failed item logged `For privacy, personal information cannot be saved to memory.`:

| User | Vendor | Outcome |
|------|--------|---------|
| 77 | chatgpt | 26 of 31 candidates saved; 5 PII-rejected |
| 224 | chatgpt | 18 of 22 saved; 4 PII-rejected (retried the same rejected candidate 3x) |
| 633 | claude + chatgpt | ~26 save batches from 11:42 to 12:40; nearly every batch `ok: 0, failed: 5`; only 4 memories ever saved. Identical batches retried 8+ times — rejection is deterministic, so every retry failed the same way |

## Root causes (validated against code)

**1. PII gate vs. import content mismatch (the "didn't work at all" user).**
`lib/nexus/memory/memory-service.ts:141-147` throws `ContentSafetyBlockedError` when `detectPII` returns *any* entity. The detector (`lib/safety/pii-detection-service.ts` → Comprehend) flags NAME (score ≥ 0.90), DATE_TIME (≥ 0.97), AGE (≥ 0.90), and EMAIL/PHONE/SSN/ADDRESS at any score (`lib/safety/types.ts:212-272`). Memory exports from ChatGPT/Claude are inherently dense with names of family/colleagues, dates, and ages, and the extraction system prompt (`lib/nexus/memory/import-prompts.ts`) excludes "contact details, sensitive identifiers" but never instructs the model to de-identify names/dates/ages. So extraction happily produces candidates the save gate is guaranteed to reject. The sanitize-at-write choke point itself is working as designed (non-negotiable invariant) — the defect is that the import pipeline feeds it content that cannot pass, then presents that as a generic failure.

**2. Rejection reason is never surfaced to the user.**
`ImportedMemoryItemResult` (`actions/nexus/memory-import.actions.ts:44-49`) carries only `status: "failed"` — no reason. The dialog (`app/(protected)/settings/_components/memory-import-dialog.tsx`) shows `"N imported; M need attention"` with no explanation, inviting blind retries of deterministic failures (user 633's hour-long retry storm).

**3. Extraction response parsing is brittle and unobservable.**
`lib/nexus/memory/memory-import.ts` calls Nova Lite (`us.amazon.nova-lite-v1:0` — confirmed as the live prod setting `MEMORY_EXTRACTION_MODEL_ID`) with forced tool choice and `maxOutputTokens: 8192`. `parseMemoryImportCandidates` does a strict whole-batch `safeParse` — one invalid candidate, an out-of-enum category, or > `MAX_MEMORY_IMPORT_CANDIDATES` (100) items rejects the entire response; a response truncated at the output-token cap yields no valid tool call at all. The failure durations (18–27.5s vs 3.3–5.7s for successes) are consistent with large outputs hitting the 8192-token cap mid-tool-call. Cannot be confirmed from logs because the error path records nothing about the model response (no finishReason, usage, output length, or toolCall presence) and not even userId/vendor/paste size. There is no retry.

## Proposed fix (in scope)

1. **Return and display per-candidate failure reasons.** Add a user-safe `reason` to `ImportedMemoryItemResult` (from `ContentSafetyBlockedError.message`; generic fallback otherwise) and render it beside each failed candidate in the import dialog so users edit instead of blind-retrying.
2. **De-identify at extraction.** Extend `MEMORY_IMPORT_EXTRACTION_SYSTEM_PROMPT` to require candidates be phrased without third-party names (use relationships/roles), specific dates, ages, or contact details — so candidates are compatible with the unchanged sanitize-at-write gate.
3. **Harden extraction parsing.** Salvage the valid subset when the batch-level schema fails (per-candidate lenient parse, capped at 100) instead of throwing away everything; retry the model call once when the response is unparseable.
4. **Make extraction failures diagnosable.** On parse failure, log finishReason, token usage, output text length, and toolCall presence (never raw content), plus vendor and paste length in the action's error context.

The PII gate in `memory-service.ts` is **not** modified.

## Definition of Done

- [ ] `saveImportedMemories` returns a user-safe `reason` for each failed candidate, and the import dialog displays it next to that candidate (unit tests cover action + dialog)
- [ ] `MEMORY_IMPORT_EXTRACTION_SYSTEM_PROMPT` instructs the model to write candidates without third-party names, specific dates, ages, or contact details (existing prompt unit test updated)
- [ ] `parseMemoryImportCandidates` returns the valid candidate subset (capped at 100) when the batch contains invalid entries or exceeds the cap, instead of throwing (unit tests)
- [ ] Extraction retries the model call exactly once when the response is unparseable, then fails with the existing error (unit test)
- [ ] Extraction parse failures log finishReason, usage, output length, and toolCall count — no raw model output or pasted content (unit test)
- [ ] PII gate behavior in `lib/nexus/memory/memory-service.ts` is unchanged (existing sanitize tests still green, no edits to the gate)
- [ ] Full test suite green
- [ ] Zero lint warnings; typecheck clean
- [ ] E2E flow(s) pass — flows: `settings-memory-import`

## Acceptance tests / E2E flows

- `settings-memory-import` (`tests/e2e/settings-memory-import.functional.spec.ts`): open Settings → Memory → Import, paste a vendor export, review extracted candidates, save; when a candidate fails, its row shows the specific reason (screenshot the failed-candidate state).

## Affected areas

- `actions/nexus/memory-import.actions.ts` — per-item failure reason
- `lib/nexus/memory/memory-import.ts` — lenient parse, retry, diagnostics
- `lib/nexus/memory/import-prompts.ts` — de-identification instructions
- `app/(protected)/settings/_components/memory-import-dialog.tsx` — reason display
- `tests/unit/lib/nexus/memory-import.test.ts`, `tests/unit/actions/nexus-memory-import-actions.test.ts`, `tests/unit/settings-memory-import-dialog.test.tsx`

## Out of scope

- Changing the PII gate policy (reject → redact) or its confidence floors — separate decision (see open decisions)
- Changing the extraction model — `MEMORY_EXTRACTION_MODEL_ID` is already settings-driven; no code change needed to switch
- Auto-extraction (#1410) prompt/pipeline beyond code it shares with import

## Open decisions (need Hagel)

1. Keep detect-and-reject, or redact PII spans (Comprehend returns offsets) for `import:*` sources? Default assumed here: keep reject, fix the pipeline around it.
2. Nova Lite retained for extraction, or bump the prod setting to a stronger model? (Settings change only — no deploy.)

## Risk & rollback

- Risk: low — additive UX field, prompt text, and parse-path hardening; no schema/migration changes; the safety gate is untouched.
- Rollback: revert the PR; no data or infra impact.

krishagel:
Closes #1610.

## What broke

Three prod users hit the Nexus memory import on 2026-08-05 between ~11:22am and ~12:40pm PDT. Two independent defects, both confirmed in `/ecs/aistudio-prod` and against the code.

**The PII gate refused the exact content memory exists to hold.** `memoryService.save` threw whenever Comprehend returned *any* entity (NAME ≥ 0.90, DATE_TIME ≥ 0.97, AGE ≥ 0.90, EMAIL/PHONE/SSN/ADDRESS at any score). A memory is the user's own record of their own life — names, relationships, dates, ages aren't incidental to that, they *are* it. Vendor exports are dense with them, so extraction reliably produced candidates the save gate was guaranteed to reject. User 633 ran ~26 save batches over an hour, nearly all `ok: 0, failed: 5`, and landed 4 memories total.

**Extraction lost every candidate when one response truncated.** The last three attempts of the day failed after 18–27.5s (vs 3.3–5.7s for the five that succeeded). The whole paste — up to 200,000 chars — went to Nova Lite in one call, strict whole-batch `safeParse`, `maxOutputTokens: 8192`, no retry. A tool call truncated mid-object parses as nothing, so the import returned zero candidates. The error path logged nothing about the response, so it wasn't diagnosable.

**And the UI never said why.** `ImportedMemoryItemResult` carried only `status: "failed"`; the dialog said "N need attention". The rejections were deterministic, so every retry failed identically.

## What changed

**PII detection is now telemetry, not a gate.** `recordPIITelemetry` logs entity types and counts — never offsets or values — and always lets the write proceed. A PII scan that *errors* no longer fails closed either. This matches the telemetry-only stance already in place for Bedrock Guardrails. The content-safety gate (`processInput`) is untouched, still blocks, and still runs before embedding or any DB access.

The prompts pulled the same way from the other end: both the import and auto-extraction prompts said to exclude "contact details, sensitive identifiers", stripping personal detail before it ever reached the gate. Both now keep names, relationships, ages, locations, dates, and contact details verbatim and exclude only credentials.

**Extraction is chunked and salvaging.**
- `splitExtractionChunks` splits on line boundaries at 12,000 chars (oversized lines split on characters), so no response approaches the output cap. Chunks run at concurrency 3, merged and deduplicated, capped at `MAX_MEMORY_IMPORT_CANDIDATES`.
- One failing chunk no longer discards the chunks that parsed.
- `parseMemoryImportCandidates` salvages every individually valid candidate when the batch schema fails, instead of discarding the response over one bad entry.
- Each chunk retries once on an unparseable response.
- Parse failures log `finishReason`, usage, output length, and tool-call count — shape only, never model output or pasted content.

**Failures explain themselves.** The action returns a user-safe `reason` per failed candidate (blocked-content messages pass through; everything else stays generic so provider/database/network internals never reach the browser). The dialog renders it on the failing row and clears it once the text is edited.

## Also fixed

`tests/e2e/settings-memory-crud.functional.spec.ts` was red before this branch — I verified that by stashing and re-running. Its first failure was the PII gate rejecting a manual memory; with that fixed it still failed at the edit step, because filling the edit textarea replaces the row's rendered text, so the `hasText` row filter matches nothing and every control inside it becomes unreachable. The spec now pins the row by `data-memory-id` before editing.

## Verification

- `bun run lint` — 0 errors, 0 warnings
- `bun run test:ci` — 571 suites / 5393 tests passing
- Authenticated E2E against live Bedrock + Comprehend (`E2E_RUN_EXTERNAL=1`) — **all 9 memory specs passing**, including a new `settings-memory-import` case that imports a wife's name, a daughter's age, a start date, and an email address and asserts all four land in the database

No schema, migration, or infrastructure changes. Rollback is a revert.

## Note on the changed decision

Issue #1610 originally proposed leaving the PII gate alone and de-identifying at extraction instead. Hagel reversed that: users need to store and import their personal information, so the gate had to go rather than the data. The gate is removed as a *refusal*; the detection signal is kept as telemetry.

### DIFF
--- a/actions/nexus/memory-import.actions.ts
+++ b/actions/nexus/memory-import.actions.ts
@@ -31,10 +31,8 @@ import {
 } from "@/lib/nexus/memory/memory-import-schemas"
 import { memoryService } from "@/lib/nexus/memory/memory-service"
 import { ContentSafetyBlockedError } from "@/lib/streaming/types"
+import { GENERIC_MEMORY_IMPORT_FAILURE_REASON } from "@/lib/nexus/memory/memory-constants"
 import { hasCapabilityAccess } from "@/utils/roles"
-
-const GENERIC_CANDIDATE_FAILURE_REASON =
-  "This memory could not be saved. Try again in a moment."
 
 interface MemoryImportRequester {
   userId: number
@@ -66,7 +64,7 @@ function candidateFailureReason(error: unknown): string {
   if (error instanceof ContentSafetyBlockedError) {
     return error.blockedMessage
   }
-  return GENERIC_CANDIDATE_FAILURE_REASON
+  return GENERIC_MEMORY_IMPORT_FAILURE_REASON
 }
 
 export interface SaveImportedMemoriesResult {
--- a/actions/nexus/memory-import.actions.ts
+++ b/actions/nexus/memory-import.actions.ts
@@ -30,7 +30,11 @@ import {
   type SaveImportedMemoriesInput,
 } from "@/lib/nexus/memory/memory-import-schemas"
 import { memoryService } from "@/lib/nexus/memory/memory-service"
+import { ContentSafetyBlockedError } from "@/lib/streaming/types"
 import { hasCapabilityAccess } from "@/utils/roles"
+
+const GENERIC_CANDIDATE_FAILURE_REASON =
+  "This memory could not be saved. Try again in a moment."
 
 interface MemoryImportRequester {
   userId: number
@@ -46,6 +50,23 @@ export interface ImportedMemoryItemResult {
   status: "saved" | "failed"
   memoryId?: string
   action?: "inserted" | "updated"
+  /** Present on failures only. Safe to render to the user. */
+  reason?: string
+}
+
+/**
+ * A user-safe explanation for one rejected candidate. Blocked-content messages
+ * are already written for users; every other failure stays generic so provider,
+ * database, and embedding internals never reach the browser.
+ *
+ * Without this the dialog showed only "N need attention", and a prod user
+ * retried the same deterministically rejected batch eight times in an hour.
+ */
+function candidateFailureReason(error: unknown): string {
+  if (error instanceof ContentSafetyBlockedError) {
+    return error.blockedMessage
+  }
+  return GENERIC_CANDIDATE_FAILURE_REASON
 }
 
 export interface SaveImportedMemoriesResult {
@@ -199,7 +220,11 @@ export async function saveImportedMemories(
             error instanceof Error ? error.message : String(error),
           ),
         })
-        results.push({ index, status: "failed" })
+        results.push({
+          index,
+          status: "failed",
+          reason: candidateFailureReason(error),
+        })
       }
     }
 
--- a/app/(protected)/settings/_components/memory-import-dialog.tsx
+++ b/app/(protected)/settings/_components/memory-import-dialog.tsx
@@ -31,6 +31,7 @@ import {
   MEMORY_IMPORT_VENDOR_GUIDES,
 } from "@/lib/nexus/memory/import-prompts"
 import {
+  GENERIC_MEMORY_IMPORT_FAILURE_REASON,
   MAX_MEMORY_IMPORT_CHARS,
   MAX_MEMORY_IMPORT_SAVE_BATCH_CANDIDATES,
   MAX_NEXUS_MEMORY_CONTENT_CHARS,
@@ -52,9 +53,6 @@ const MEMORY_CATEGORIES: ReadonlyArray<{
   { value: "preference", label: "Preference" },
   { value: "context", label: "Context" },
 ]
-
-const GENERIC_CANDIDATE_FAILURE_REASON =
-  "This memory could not be saved. Try again in a moment."
 
 interface CandidateDraft extends MemoryImportCandidate {
   id: string
@@ -130,7 +128,7 @@ async function saveCandidateBatches(
       if (draftIndex !== undefined) {
         failureReasons.set(
           draftIndex,
-          item.reason ?? GENERIC_CANDIDATE_FAILURE_REASON,
+          item.reason ?? GENERIC_MEMORY_IMPORT_FAILURE_REASON,
         )
       }
     }
--- a/app/(protected)/settings/_components/memory-import-dialog.tsx
+++ b/app/(protected)/settings/_components/memory-import-dialog.tsx
@@ -53,9 +53,14 @@ const MEMORY_CATEGORIES: ReadonlyArray<{
   { value: "context", label: "Context" },
 ]
 
+const GENERIC_CANDIDATE_FAILURE_REASON =
+  "This memory could not be saved. Try again in a moment."
+
 interface CandidateDraft extends MemoryImportCandidate {
   id: string
   selected: boolean
+  /** Why the last save attempt rejected this candidate, if it did. */
+  failureReason?: string
 }
 
 interface SaveProgress {
@@ -70,7 +75,8 @@ interface SelectedCandidate {
 
 interface BatchedSaveOutcome {
   successful: number
-  failedDraftIndexes: Set<number>
+  /** Draft index -> user-safe reason, so each failed row can explain itself. */
+  failureReasons: Map<number, string>
   requestFailureMessage: string | null
 }
 
@@ -91,7 +97,7 @@ async function saveCandidateBatches(
 ): Promise<BatchedSaveOutcome> {
   let successful = 0
   let requestFailureMessage: string | null = null
-  const failedDraftIndexes = new Set<number>()
+  const failureReasons = new Map<number, string>()
 
   for (
     let offset = 0;
@@ -112,7 +118,7 @@ async function saveCandidateBatches(
     if (!result.isSuccess) {
       requestFailureMessage = result.message
       for (const { draftIndex } of selected.slice(offset)) {
-        failedDraftIndexes.add(draftIndex)
+        failureReasons.set(draftIndex, result.message)
       }
       break
     }
@@ -122,7 +128,10 @@ async function saveCandidateBatches(
       if (item.status !== "failed") continue
       const draftIndex = batch[item.index]?.draftIndex
       if (draftIndex !== undefined) {
-        failedDraftIndexes.add(draftIndex)
+        failureReasons.set(
+          draftIndex,
+          item.reason ?? GENERIC_CANDIDATE_FAILURE_REASON,
+        )
       }
     }
     onProgress({
@@ -133,7 +142,7 @@ async function saveCandidateBatches(
 
   return {
     successful,
-    failedDraftIndexes,
+    failureReasons,
     requestFailureMessage,
   }
 }
@@ -275,6 +284,87 @@ function ImportSourceStep({
   )
 }
 
+type CandidateUpdate = Partial<
+  Pick<CandidateDraft, "selected" | "content" | "category">
+>
+
+function ImportCandidateRow({
+  candidate,
+  index,
+  isSaving,
+  onUpdate,
+}: {
+  candidate: CandidateDraft
+  index: number
+  isSaving: boolean
+  onUpdate: (id: string, update: CandidateUpdate) => void
+}) {
+  return (
+    <div
+      data-testid="memory-import-candidate"
+      className={`space-y-3 rounded-md border p-3 ${
+        candidate.failureReason ? "border-destructive/50" : ""
+      }`}
+    >
+      {candidate.failureReason && (
+        <p
+          role="alert"
+          data-testid={`memory-import-candidate-reason-${index}`}
+          className="text-sm text-destructive"
+        >
+          {candidate.failureReason}
+        </p>
+      )}
+      <div className="flex items-start gap-3">
+        <Checkbox
+          checked={candidate.selected}
+          disabled={isSaving}
+          data-testid={`memory-import-candidate-select-${index}`}
+          aria-label={`Import candidate ${index + 1}`}
+          className="mt-2"
+          onCheckedChange={(checked) =>
+            onUpdate(candidate.id, { selected: checked === true })
+          }
+        />
+        <Textarea
+          value={candidate.content}
+          maxLength={MAX_NEXUS_MEMORY_CONTENT_CHARS}
+          disabled={isSaving}
+          data-testid={`memory-import-candidate-content-${index}`}
+          aria-label={`Memory candidate ${index + 1} content`}
+          onChange={(event) =>
+            onUpdate(candidate.id, { content: event.target.value })
+          }
+        />
+      </div>
+      <Select
+        value={candidate.category}
+        disabled={isSaving}
+        onValueChange={(value) =>
+          onUpdate(candidate.id, {
+            category: value as NexusMemoryCategory,
+          })
+        }
+      >
+        <SelectTrigger
+          className="w-full sm:w-44"
+          data-testid={`memory-import-candidate-category-${index}`}
+          aria-label={`Memory candidate ${index + 1} category`}
+        >
+          <SelectValue />
+        </SelectTrigger>
+        <SelectContent className={meridianPortalClassName}>
+          {MEMORY_CATEGORIES.map((category) => (
+            <SelectItem key={category.value} value={category.value}>
+              {category.label}
+            </SelectItem>
+          ))}
+        </SelectContent>
+      </Select>
+    </div>
+  )
+}
+
 function ImportReviewStep({
   candidates,
   isSaving,
@@ -293,15 +383,21 @@ function ImportReviewStep({
   const selectedCount = candidates.filter(
     (candidate) => candidate.selected,
   ).length
-  const updateCandidate = (
-    id: string,
-    update: Partial<
-      Pick<CandidateDraft, "selected" | "content" | "category">
-    >,
-  ) => {
+  const updateCandidate = (id: string, update: CandidateUpdate) => {
     onCandidatesChange(
       candidates.map((candidate) =>
-        candidate.id === id ? { ...candidate, ...update } : candidate,
+        candidate.id === id
+          ? {
+              ...candidate,
+              ...update,
+              // An edit answers the rejection, so the old reason no longer
+              // describes the text on screen.
+              failureReason:
+                update.content === undefined
+                  ? candidate.failureReason
+                  : undefined,
+            }
+          : candidate,
       ),
     )
   }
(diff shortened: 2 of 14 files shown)


krishagel:
Pushed 8d4e85a7e addressing both reviews.

**Codex P1 — accepted, and it caught real over-reach.** `scheduleNexusMemoryAutoExtraction` does fire after every persisted turn (verified at `app/api/nexus/chat/route.ts:208-218`), unattended and with nothing to review. Loosening *that* prompt was not part of the reported problem and went beyond what was asked for, so it is reverted to the original exclusion of contact details and sensitive identifiers.

The distinction that survives: import is the user pasting their own export and approving each candidate by hand, so it keeps the personal detail. Auto-extraction is unattended, so it does not. Neither path *refuses* a write — the gate is gone from both — the difference is only in what the model is asked to pull out.

**Docs were also wrong, and that's now fixed.** `docs/features/k12-content-safety.md` claimed "relevant K-12 PII and district-specific identifiers are refused. Detection errors fail closed," repeated in the COPPA section and the PII-mapping summary. All three now describe the real behavior and which path still holds third-party identifiers back.

**Claude's truncation point — accepted, this was the important one.** Retrying an identical prompt at `temperature: 0` cannot fix a `finishReason: "length"` response, which is the exact failure class the incident was about. `extractChunk` now halves an over-budget chunk and extracts the halves (bounded to 3 splits, floor 500 chars). A failure that is *not* an overrun still just retries and does not fan out. Both paths have tests.

Also taken: `runExtraction` moved inside the try so throttled Bedrock calls get the retry; cap drops are logged in both `salvageCandidates` and `dedupeCandidates`; the duplicated failure-reason string now lives once in `memory-constants`.

**One flake I introduced, fixed.** The new PII E2E case asserted exactly 4 extracted candidates — it failed once and passed on retry. Candidate count is the model's call, so the test now asserts a floor, deselects any surplus, and pins the saved set itself. Three consecutive clean runs, no retries.

Gate: lint 0/0, typecheck 0, 5397 unit tests, all 9 memory E2E specs green with no retries.

krishagel:
Pushed a8b55a204 with the two nits from the second round.

**Jittered backoff before a retry.** Immediate retry at concurrency 3 tends to land in the same Bedrock throttle window that just rejected it. Now waits 250ms × attempt plus equal jitter, skipped on the output-budget path since that splits rather than retries.

**`droppedCandidates` no longer overstates.** It counted schema-invalid entries past the cap as cap losses; validation now happens before the cap check.

**Consciously accepting the recursive-split call ceiling rather than adding a budget.** The bound already exists: 2 attempts, then at most 3 levels of halving → `2 + 2(2 + 2(2 + 2×2)) = 30` calls for one top-level chunk. Reaching it requires every sub-chunk to overrun the output budget at every depth, down to 1,500 characters — and 1,500 characters cannot produce 8,192 output tokens of candidate JSON when each candidate is capped at 500 chars. A global call budget would mean a paste silently yielding fewer memories, which is the unasked-for-cap class this repo avoids.

**On the trust-boundary question** — you're right that it's a deliberate policy call, and it's the owner's, not mine. The posture is: import is attended and user-approved, auto-extraction is unattended and keeps the stricter prompt. I've flagged the long-term-posture question directly to Hagel rather than deciding it here.

Gate: lint 0/0, typecheck 0, 5397 unit tests, all 9 memory E2E specs green with no retries.

krishagel:
Pushed f1a4c44bd. That was a real bug, and my own comment was lying about it.

**Overrun now breaks out of the attempt loop.** The comment said the split path skipped the retry, but only the *delay* was skipped — the loop still ran a second identical call at temperature 0 before falling through to `extractSplitHalves`. A guaranteed-identical truncation, paid for in latency and spend on every chunk that overran, compounding at each split depth. Worst case per top-level chunk drops from 30 model calls to `1 + 2(1 + 2(1 + 2×2)) = 15`.

**The test that should have caught it now does.** The split case asserted recovered candidates but never call counts — exactly why the waste went unnoticed. It now asserts 3 calls (one overrun, then the two halves), plus a new case for a chunk too small to split: one call, no retry, no fan-out. And `logs response shape and fails after the retry is exhausted` was using `finishReason: "length"` to drive exhaustion, which is no longer a retrying path at all — it now uses a `stop` response so it tests the path it names.

**On the `recordPIITelemetry` nit — checked, and it's moot today.** Every return in `ContentSafetyService.processInput` sets `processedContent` to the input unchanged (`lib/safety/content-safety-service.ts:74, 86, 138, 150, 163, 175`), so the two are identical. Leaving it scanning the submitted content, which stays the more faithful telemetry if that ever changes.

Gate: lint 0/0, typecheck 0, 5398 unit tests, all 9 memory E2E specs green with no retries.

krishagel:
Pushed ca101932a — the concurrency finding was correct.

**Split recovery now stays inside the declared bound.** `extractSplitHalves` fanned its halves out through its own `Promise.allSettled`, invisible to the top-level worker pool. Since recursion happens *inside* a worker slot, parallel halves at max depth could put ~8 calls in flight per slot and ~24 overall — 8× `MEMORY_EXTRACTION_MAX_CONCURRENCY`, and exactly the pile-up the retry jitter exists to prevent. The halves now run sequentially: at most two per level, rare path, and one in-flight call per slot makes the bound actually true. Partial recovery is unchanged.

**Worker-pool comment added**, explaining why `nextChunk++` needs no lock (synchronous until the first await, so the index is claimed before any other worker can be scheduled).

Two left as-is, deliberately:
- **Double cap warning** — the two are distinct events with distinct messages (per-chunk batch vs global merge) and both are true, so the pair is diagnosable rather than misleading.
- **Sequential saves in `saveImportedMemories`** — you flagged it as not a regression from this PR, and it isn't. Bounded at 5 candidates per batch. Folding it in would widen an incident fix into a latency refactor.

On compliance sign-off: correct to flag it. The decision is the platform owner's and he made it explicitly; I've also raised the remaining trust-boundary question (import carries third-party PII with only the user's review step gating it, unlike auto-extraction) directly to him rather than resolving it in code.

Gate: lint 0/0, typecheck 0, 5398 unit tests, all 9 memory E2E specs green.

krishagel:
Pushed b0fc7ac1a — took both observability notes.

**One cost-summary line per import.** Squarely the spirit of this incident: the 2026-08-05 failures were undiagnosable because the error path recorded nothing. Per-chunk warnings fixed the failure case, but total model spend still had to be reassembled by counting warnings. `extractMemoryImportCandidates` now threads an `ExtractionStats` counter through the chunk context and logs vendor, pasted length, chunk count, model calls, split events, failed chunks, and candidate count. Asserted in the split test (3 model calls, 1 split event).

**`recordPIITelemetry` reasoning now lives in the code.** It scans the submitted content rather than `safety.processedContent`; every return in `ContentSafetyService.processInput` passes content through unchanged (`lib/safety/content-safety-service.ts:74, 86, 138, 150, 163, 175`), so they're identical today, and if that ever changes the submitted text is the more faithful thing to report. Raised in two rounds, so it belongs next to the code rather than in a commit message.

Across five rounds the findings have gone P1 scope error → wasted retry call → concurrency-bound escape → observability polish, which reads like convergence. Everything flagged has been fixed or answered with reasoning.

Gate each round: lint 0/0, typecheck 0, 5398 unit tests, all 9 memory E2E specs green with no retries.

krishagel:
Pushed 19c460d52.

**Defense-in-depth point taken, and made observable rather than deferred.** You're right that a prompt is a weaker control than the deterministic gate it replaced — if the extraction model drifts on instruction-following, the auto path has no second line. Rather than file a follow-up, `recordPIITelemetry` now tags every detection with the write's `source`. A `source:"auto"` line carrying NAME or AGE is precisely the signal that the prompt didn't hold, and it's alertable today. Unit test asserts the source reaches the log.

**Real email removed from test data.** `hagelk@psd401.net` sat in test source, CI output, and every Comprehend call the spec made. A fabricated address trips EMAIL detection identically, so the fixture is synthetic throughout now. Good catch — no test value was being bought with that.

**Sequential saves: leaving as-is**, third time it's come up. Pre-existing, bounded at 5 per request, and as you note sequential is arguably correct against hammering the embedding provider. Changing it turns an incident fix into a latency refactor.

Gate: lint 0/0, typecheck 0, 5399 unit tests, all 9 memory E2E specs green with no retries.

krishagel:
Pushed 4b6ace35c. That was a real miss and the right thing to block on.

I'd updated three places in `k12-content-safety.md` and left the rest of the same file asserting the opposite. For a COPPA/FERPA doc that's worse than silence — a reader hits both claims and can't tell which is current.

Fixed all four you named (overview bullet, flow diagram, architecture ASCII diagram, and the "what if the safety service is unavailable" answer claiming memory writes fail closed), then swept the rest of the file for the same drift rather than leaving the terminology half-converted:

- Headings and prose calling these "gates" now say detection/telemetry — they gate nothing, they observe
- Custom-patterns note now states they do not refuse a write
- The feature-toggle note no longer calls them "explicit application safety boundaries"; they're observability with nothing to toggle off, and it now points at `processInput` as the actual boundary
- The observability bullet notes the memory event carries the write's `source`, so unattended auto-extraction is queryable on its own

`processInput` is now named throughout as the one hard gate that still refuses a durable write.

Documentation only — no code, schema, or infrastructure change.

One thing worth recording rather than hiding: a single full-suite run reported one failing suite without naming it, on this documentation-only diff. It didn't reproduce across three further runs (exit 0, 571 suites each), and a markdown edit can't change test outcomes. Flagging it instead of quietly re-running to green.

Gate: lint 0/0, typecheck 0, 571 suites passing.

krishagel:
Pushed b64793d45 — took the two split-path notes, and I'm stopping here on polish.

**The partial-recovery guarantee had no test.** `extractSplitHalves` promises one failing half doesn't cost the half that parsed — the whole reason it doesn't rethrow — and nothing exercised that at the top level. That's exactly how it quietly regresses into "one bad half loses the chunk" with no test going red. New case: chunk overruns, splits, one half fails both attempts, other succeeds; asserts the surviving candidates come back plus the exact call count (1 overrun + 2 doomed + 1 good).

**A lost half is no longer silent.** The catch swallowed the error with no per-half record, so the only trace was a smaller `candidateCount` — enough to notice something vanished, not enough to say what. Now logs vendor, chunk index, split depth, half index, and half size.

Consciously not taking, all previously reasoned through: the double cap-warning (distinct events, distinct messages, both true), sequential saves in `saveImportedMemories` (pre-existing, bounded at 5/request), and the prompt-adherence dependency on auto-extraction — which is now *observable* rather than merely flagged, since `recordPIITelemetry` tags detections with the write's `source`.

Gate: lint 0/0, typecheck 0, 5400 unit tests, all 9 memory E2E specs green with no retries. CI has passed every round.

The remaining open items are policy, not code: whether import's user-review-only trust boundary is the intended long-term posture for K-12, and whether the compliance record for removing the refusal is sufficient. Both are the platform owner's calls and are with him.

PR #1613, opened 2026-08-06
fix(agent): let the agent read its own schedule run history, and quiet routine contention

krishagel:
Both items from the post-deploy prod review on 2026-08-06.

## 1. The agent can now diagnose its own scheduled-run failures

zarlingj's "Nightly inbox filing" failed, they asked the agent how to fix it, and it replied:

> *"This session has no direct access to that scheduled job's run logs to diagnose the specific error."* — `agent_failures` 2580

An ironic gap: the failed-turn message had just been made self-describing (*"I had already run: read, write, exec, edit… exec was still running and may or may not have completed"*), and the agent still couldn't read the history needed to act on it. The user got an apology instead of an answer.

`list` was the only window onto runs, and it reports a **status without a reason**. Added a `runs` operation end to end — reader → service → broker route → `runs.js` skill command — returning recent runs newest-first with `status`, `errorMessage`, `latencyMs` and `createdAt`, optionally narrowed to one schedule.

**Scoping:** owner-scoped at the query like every other operation on this service, so a caller can't read another owner's runs by passing their `scheduleId`. The owner email is normalised *before* it reaches the reader, so case can't be used to sidestep that. Bounded to 50 rows — this lands in an agent's context window, and an owner with a five-minute schedule accumulates thousands.

**Documented, not just built.** SKILL.md carries the instruction that actually matters:

> Do NOT tell the user you have no access to the scheduled job's logs — you do. Read the runs, quote the error, and say what you will change.

Allowlisting without documenting leaves a command inert; that lesson cost us a round on `accessproposals` in #1608.

## 2. Cross-schedule lock contention is `warn`, not `error`

Every schedule for one owner shares the workspace lock, so two firing close together contend **by design** and the retry resolves it. I downgraded only the *same-fire* case in #1607 — this is the sibling branch, still putting routine self-healing contention in the operator feed beside real outages.

## Verification

- Full unit suite: **571 suites / 5,421 tests passing**

Typecheck caught the two service test doubles that needed the new reader method — worth noting as evidence the interface is load-bearing rather than optional.

New coverage: error messages surface, owner normalisation reaches the reader, narrowing args pass through, and non-string `scheduleId` / non-numeric `limit` are ignored rather than forwarded.

## Context — where the prod baseline stands

After the two images shipped today, prod went from **119 failures / 35 users over three days** to **10 in ~17 hours**, with every top class at zero: `ChatDeadlineExpired*`, `phase1-forbidden` self-reports, Slides API, "no SIS data tool", and `ChatResponseDeliveryFailed`.

Still open and untouched, so they don't read as regressions here: `EmptyAgentResponse` (4 rows, 1 user, overnight), `OpenClawIncompleteToolTurn` / `OpenClawChatError` (message improved, frequency unchanged), and `students_demographics` timeouts (warehouse-side, with the data engineer).

## Checklist

- [x] No `console.*` in production code
- [x] Typecheck passes
- [x] All tests pass
- [x] No `any` types; no new type assertions
- [x] Documentation updated — psd-schedules SKILL.md

### DIFF
--- a/app/api/agent/schedules/route.ts
+++ b/app/api/agent/schedules/route.ts
@@ -76,13 +76,20 @@ async function executeScheduleOperation(
           enabled: body.enabled,
         }),
       })
+    case "runs":
+      return NextResponse.json({
+        runs: await service.runs(ownerEmail, {
+          scheduleId: body.scheduleId,
+          limit: body.limit,
+        }),
+      })
     case "delete":
       return NextResponse.json({
         deleted: await service.delete(ownerEmail, body.scheduleId),
       })
     default:
       return NextResponse.json(
-        { error: "operation must be create, list, update, or delete" },
+        { error: "operation must be create, list, runs, update, or delete" },
         { status: 400 }
       )
   }
--- a/infra/agent-image/skills/psd-schedules/SKILL.md
+++ b/infra/agent-image/skills/psd-schedules/SKILL.md
@@ -165,3 +165,23 @@ and confirm. Run `create`. Tell the user when the first fire will be.
   do not try to "fix" unrelated infrastructure (gateway pairing, device
   approval, etc.) — those are signs of a wrong tool choice, not a real
   problem to solve.
+
+## Diagnosing a failed run
+
+`list` reports each schedule's last status but never the reason. When a user
+asks why a scheduled job failed, read the run history — it carries the error
+message:
+
+```bash
+node /opt/psd-skills/psd-schedules/runs.js
+node /opt/psd-skills/psd-schedules/runs.js --schedule-id <id> --limit 10
+```
+
+Newest first, up to 50, scoped to the caller's own schedules. Each run reports
+`status`, `errorMessage`, `latencyMs` and `createdAt`.
+
+Do NOT tell the user you have no access to the scheduled job's logs — you do.
+Read the runs, quote the error, and say what you will change. Observed
+2026-08-06: a nightly filing job failed, the user asked how to fix it, and the
+agent replied that it had no way to see that job's run logs.
+
--- a/infra/agent-image/skills/psd-schedules/authority.test.js
+++ b/infra/agent-image/skills/psd-schedules/authority.test.js
@@ -7,7 +7,7 @@ const path = require('node:path');
 const skillDir = __dirname;
 
 describe('psd-schedules authority selectors', () => {
-  for (const script of ['create.js', 'list.js', 'update.js', 'delete.js']) {
+  for (const script of ['create.js', 'list.js', 'update.js', 'delete.js', 'runs.js']) {
     it(`${script} rejects the legacy --user selector before network access`, () => {
       const result = spawnSync(
         process.execPath,
--- a/infra/agent-image/skills/psd-schedules/runs.js
+++ b/infra/agent-image/skills/psd-schedules/runs.js
@@ -24,9 +24,15 @@ async function main() {
   }
   rejectLegacyAuthorityArgs(args);
 
+  const scheduleId = args['schedule-id'];
+  if (scheduleId === true) fail('--schedule-id requires a value');
+
   const limitArg = args.limit;
   let limit;
   if (limitArg !== undefined) {
+    // A valueless --limit parses as `true`, and Number(true) is 1 — without this
+    // guard a malformed invocation silently becomes `--limit 1`.
+    if (limitArg === true) fail('--limit requires a value');
     limit = Number(limitArg);
     if (!Number.isInteger(limit) || limit < 1 || limit > 50) {
       fail('--limit must be an integer between 1 and 50');
@@ -35,7 +41,7 @@ async function main() {
 
   const result = await requestScheduleOperation({
     operation: 'runs',
-    ...(args['schedule-id'] ? { scheduleId: args['schedule-id'] } : {}),
+    ...(scheduleId ? { scheduleId } : {}),
     ...(limit === undefined ? {} : { limit }),
   });
   emit(result);
--- a/infra/agent-image/skills/psd-schedules/runs.js
+++ b/infra/agent-image/skills/psd-schedules/runs.js
@@ -1,0 +1,48 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const {
+  fail,
+  rejectLegacyAuthorityArgs,
+  parseArgs,
+  emit,
+  requestScheduleOperation,
+} = require('./common');
+
+async function main() {
+  const args = parseArgs(process.argv);
+  if (args.help) {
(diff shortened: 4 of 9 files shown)


krishagel:
@codex review

Head is `8d040c4101`. Your P2 is fixed.

Fractional `limit` values reached Drizzle's `.limit()` because `runs()` accepted anything finite and the reader's clamp preserves fractions — `Math.min`/`Math.max` don't round. `runs.js` already required an integer, but the broker route is reachable owner-mode without the skill, so both layers now enforce it: the service requires `Number.isInteger`, and the reader floors as a last resort immediately before the query. Covered for `1.5`, `NaN` and `Infinity`, each falling back to the default rather than being forwarded.

For context on the earlier `claude-review` failure on `ce7d81c1` — that was a GitHub Actions outage (`Failed to resolve action download info: Service Unavailable`), not a finding. It passes on this SHA.

Verified locally: full unit suite 5422 tests, lint clean at `--max-warnings 0`, typecheck clean, E2E green via the pre-push gate.

krishagel:
🤖 pr-fix routine round 1 — all four review items addressed. Head is now `9a34de6f`.

### Should fix — `authority.test.js` missed `runs.js` ✅

Confirmed as an oversight, not an intentional exclusion. `authority.test.js:10` looped `[create.js, list.js, update.js, delete.js]`; `runs.js` calls `rejectLegacyAuthorityArgs(args)` exactly like its siblings but was never added, so the guard on the new command was unverified by the suite that exists specifically to pin it.

Added to the loop — the suite goes 4 scripts → 5, and `runs.js` passes with no source change, which confirms the guard itself was already wired correctly. **25/25 pass** across the psd-schedules skill (`bun run test:skill:schedules`).

### Nit — valueless `--limit` silently became `--limit 1` ✅

Fixed, and it was worth fixing: `Number(true)` is `1`, which satisfies `Number.isInteger(limit) && limit >= 1 && limit <= 50`, so the malformed invocation passed validation rather than failing. `--schedule-id` with no value forwarded `scheduleId: true`, which the service drops to `undefined` — safe, but it reads at the CLI as though narrowing was applied when it was not.

Both now `fail()` before any network access, matching the `=== true` convention `create.js` / `update.js` / `delete.js` already follow. Verified directly:

```
$ node runs.js --limit         → psd-schedules: --limit requires a value        (exit 1)
$ node runs.js --schedule-id   → psd-schedules: --schedule-id requires a value  (exit 1)
$ node runs.js --limit 999     → --limit must be an integer between 1 and 50    (unchanged)
$ node runs.js --help          → unchanged
```

### Nits — test formatting ✅

`latestBySchedule` re-indented to line up with `recentForOwner`, and the fractional-limit test moved inside `describe("schedule run history")` so all four run-history tests report as one group.

### On the indexing note

Not actioned, deliberately — you flagged it as fine as-is, and the second review confirmed `idx_agent_scheduled_runs_user (user_id, created_at desc)` already backs the primary query path. Recording it here rather than opening a follow-up, since the 50-row cap bounds the cost.

### Verification caveat

`bun test` for the skill scripts runs dependency-free and passed here. Jest, lint and typecheck could **not** be run in this routine's sandbox — `node_modules` is absent and `registry.npmjs.org` returns 403 under the egress policy, so `bun install` cannot complete. Those three are covered by the **Test, Lint, and Type Check** CI job on this SHA; I have not marked anything green that I did not actually observe. The TS edit is whitespace plus a block move, verified to parse via Bun's transpiler, and both JS files pass `node --check`.



---

krishagel:
@codex review

Head is `9a34de6fa3` — three further commits since your last pass, all from @claude's review:

- **`445c4d45`** — `authority.test.js` never covered `runs.js`. A real oversight: the suite asserts every command rejects the legacy authority args, and a new command that silently skipped it would be exactly how that boundary erodes.
- **`5fefd08e`** — `runs.js` treated a valueless `--schedule-id` / `--limit` as `true` rather than failing. `parseArgs` returns `true` for a flag with no value, so `--schedule-id` alone would have been forwarded as a boolean.
- **`9a34de6f`** — test formatting only.

And from your P2 earlier, `8d040c41`: fractional `limit` now rejected at the service (`Number.isInteger`) and floored at the reader immediately before Drizzle's `.limit()`.

Verified independently after pulling: full unit suite 5422 tests, psd-schedules skill suites 25, lint clean at `--max-warnings 0`, typecheck clean.

Note on the earlier `claude-review` failure on `ce7d81c1` — GitHub Actions outage (`Failed to resolve action download info: Service Unavailable`), not a finding.

krishagel:
🤖 pr-fix routine — **all 7 checks green on `9a34de6f`**, `mergeable_state: clean`. Labeling `pr-fix-done`.

| Check | Result |
|---|---|
| Test, Lint, and Type Check | ✅ |
| Unified Content PostgreSQL Lifecycle | ✅ |
| Validate CDK Infrastructure | ✅ |
| Auth Edge Production Artifact | ✅ |
| Analyze (javascript-typescript) | ✅ |
| CodeQL | ✅ |
| claude-review | ✅ |

**Test, Lint, and Type Check passing is the load-bearing one here** — it's what actually verified the jest/lint/typecheck I flagged as unrunnable in my sandbox, so that caveat is now closed rather than merely outstanding.

### The red builds along the way were not this branch

Three jobs failed on first run and I want the record clear, because this is the second time today this signature has shown up on this PR. All three died at `Prepare all required actions` with `Failed to resolve action download info: Service Unavailable` — a GitHub Actions control-plane outage, before a runner ever checked out the repo or executed a line of project code. Same signature @krishagel already identified on `ce7d81c1`. A fourth job then sat queued ~15 min and was cancelled for want of a runner — outage aftermath, not a test result.

Re-running cleared all four with no code change, which is itself the proof they were infrastructural.

### Two optional notes from the latest review — deliberately not actioned

Both were marked non-blocking by the reviewer, and I'm leaving them rather than quietly folding them in:

1. **`AgentScheduleLastRun` / `AgentScheduleRun` duplicate `createdAt` / `status` / `errorMessage`.** Called "purely a maintenance nit, not a correctness issue." Composing the types touches a shared interface with two known implementors, which is a wider blast radius than a green PR warrants at this point.
2. **`runs.js` doesn't trim `--schedule-id` client-side.** Called "cosmetic only" — the service trims server-side, so behaviour is already correct.

My reasoning for stopping: each additional commit restarts a ~20-minute CI cycle, and doing that for cosmetics *during an Actions outage window* trades real risk for no behavioural gain. Both are genuinely trivial if you want them — say the word and the next fire will pick them up.

### Not actioned, flagged for you

The unchecked **"Code reviewed and approved"** box is a human approval and not something this routine can or should tick. Everything mechanical is done; the PR is waiting on you.



---

PR #1616, opened 2026-08-07
fix(agent): exclude AGENTS.md from workspace sync so the rules survive cold start

krishagel:
Caught on dev immediately after deploying the bootstrap split in #1614. **The dev image currently ships with no operating rules in the system prompt.**

Asked to quote Rule 4 and Rule 9 without loading anything, the agent replied *"the psd-rules skill content isn't loaded into my context right now."*

## The split was fine — the file was being deleted

Verified in the built image:
- `/home/node/.openclaw/AGENTS.md` is written correctly — **21,915 bytes**
- The pinned host recognises the filename — **90 references** in `/app/dist`, more than `SOUL.md`'s 39

The file simply wasn't there by the time the prompt was assembled.

`workspace_sync` pulls each owner's workspace from S3 on cold start. Files owned by the **image** rather than the **user** must be excluded from that sync or the pull overwrites them. `SOUL.md` is on that list; `AGENTS.md` was not. Every existing user's S3 workspace predates `AGENTS.md`, so the pull replaced the workspace without it.

This is the same bug the exclusion list's own comment documents from **2026-04-22**, when an old `SOUL.md` in each user's workspace kept overwriting the image's fresh one and new rules *"never seemed to take"*. Identical shape, identical cause, one file later.

## Why it hid so well

An image-owned file that isn't excluded works **perfectly for a brand-new user** and is **silently absent for everyone else**. Two of the three post-deploy checks passed for exactly that reason — they don't depend on `AGENTS.md`. Only the check written specifically to detect this caught it.

## The fix

Add `AGENTS.md` to `checkpointExclusions.relativePrefixes` in `workspace-policy.json` — read by **both** the image runtime and the web checkpoint, so the two can't drift into different definitions of durable state.

Nothing moved. The rules stay in `AGENTS.md`; the sync layer now leaves them alone, exactly as it does for `SOUL.md`.

| Kind | Files | Behavior |
|---|---|---|
| Image-owned | `SOUL.md`, **`AGENTS.md`**, `openclaw.json` | Never synced — image copy always wins |
| User-owned | `IDENTITY.md`, `USER.md`, `MEMORY.md` | Must round-trip — agent-written |

`AGENTS.md` was in neither bucket. That's the whole bug.

## Tests — mutation-checked

Removing `AGENTS.md` from the policy fails two of the three:

- every image-owned bootstrap file is excluded from sync
- `IDENTITY`/`USER`/`MEMORY` are **not** excluded — those are agent-written and must round-trip; excluding them would stop the agent's memory persisting
- the generated bootstrap filename is read **out of the Dockerfile** rather than hardcoded, so renaming it fails the test instead of silently leaving the new name unprotected

## Verification

- Full unit suite; agent-image `workspace_sync` **84 tests**
- bootstrap-budget and config-consistency gates

## Checklist

- [x] No `console.*` in production code
- [x] Typecheck passes
- [x] All tests pass

### DIFF
--- a/infra/agent-image/test_workspace_sync.py
+++ b/infra/agent-image/test_workspace_sync.py
@@ -3331,11 +3331,27 @@ class ImageOwnedBootstrapExclusionTests(unittest.TestCase):
                 "stop the agent's own memory from persisting",
             )
 
-    def test_the_rules_file_the_dockerfile_writes_is_covered(self):
-        # Ties the exclusion to the Dockerfile rather than to a literal: if the
+    def test_every_bootstrap_file_the_dockerfile_generates_is_covered(self):
+        # Ties the exclusion to the Dockerfile rather than to a literal: if a
         # generated bootstrap filename changes, this fails instead of silently
-        # leaving the new name unexcluded.
+        # leaving the new name unprotected.
+        #
+        # EVERY redirect target, not the first. `re.search` returned SOUL.md —
+        # which is excluded — so the test passed while asserting nothing about
+        # AGENTS.md, the very file it was written to guard. Renaming AGENTS.md
+        # would not have failed it.
         dockerfile = pathlib.Path(__file__).with_name("Dockerfile").read_text()
-        match = re.search(r"> /home/node/\.openclaw/([A-Z]+\.md)", dockerfile)
-        self.assertIsNotNone(match, "no generated bootstrap file found in the Dockerfile")
-        self.assertIn(match.group(1), workspace_sync._SKIP_RELATIVE_PREFIXES)
+        generated = set(
+            re.findall(r"> /home/node/\.openclaw/([A-Za-z]+\.md)", dockerfile)
+        )
+        self.assertTrue(generated, "no generated bootstrap file found in the Dockerfile")
+        # Both files the image generates today; a new one must be added here
+        # AND to the policy, which is the point of the check.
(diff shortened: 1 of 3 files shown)


krishagel:
@codex review

Head is `021c22de1d`. Your P2 was correct and it was a real defect in the test I had just written to guard this.

`re.search` returns the FIRST match, and the Dockerfile redirects `SOUL.md` before `AGENTS.md` — so the test resolved to `SOUL.md`, which is excluded, and passed while asserting **nothing** about `AGENTS.md`. Renaming `AGENTS.md`, the exact regression named in its own docstring, would not have failed it.

My earlier mutation check missed it because removing `AGENTS.md` from the policy failed the other two tests in the class — the suite went red for the right reason and hid that this third test was inert. A passing suite is not evidence a specific test works.

Now uses `re.findall` over every redirect target, pins the expected set (`{SOUL.md, AGENTS.md}`) so a newly generated bootstrap file has to be added deliberately rather than silently ignored, and asserts each one is excluded. Re-mutation-checked by renaming `AGENTS.md` to `RULES.md` in the Dockerfile: **1 failure**, clean again once restored.

Note on the earlier E2E push block: `bun run test:e2e:local` failed once and passed on rerun at 319. This diff is three files — two Python, one JSON policy — none of which E2E touches.

PR #1619, opened 2026-08-07
fix(agent): put the operating rules back in SOUL.md and raise the per-file cap

krishagel:
Reverts the `AGENTS.md` split from #1614/#1616. **It was built on a wrong premise and left dev unable to restore a workspace.**

## AGENTS.md is not a bootstrap file in the host we pin

Verified three ways, all authoritative:

- `openclaw config schema` — `skipOptionalBootstrapFiles` enumerates **SOUL.md, USER.md, IDENTITY.md, HEARTBEAT.md**
- The injected-file arrays in `/app/dist` are `["SOUL.md"]` and `["IDENTITY.md","USER.md","SOUL.md"]`
- AGENTS.md is handled in `selection-*.js` under `DEFAULT_POST_COMPACTION_SECTIONS = ["Session Startup", "Red Lines"]`

This host uses AGENTS.md for **post-compaction re-injection of named sections**, not as a per-turn system prompt. Our rules file has no such sections, so **nothing was ever injected**.

## What actually broke

The rules were absent from the prompt the moment they moved. The sync exclusion added to "fix" that fixed nothing — it retired a path already present in S3, and the broker correctly refused every restore:

> `Retired workspace host state requires controlled migration`

Dev couldn't start a turn.

**The reasoning error:** I cited "90 references to AGENTS.md in `/app/dist`" as evidence it was supported. That proved only that the host *mentions the name*, for a different feature. The claim that AGENTS.md is an operational-rules bootstrap file came from a **blog post** and was never checked against the pinned host — the same mistake as `bootstrapPromptTruncationWarning` an hour earlier, which this host also rejects.

## The change

- **Dockerfile** appends the `psd-rules` body to `SOUL.md` again — a real bootstrap file, already sync-excluded, demonstrably works.
- **`workspace-policy.json`** drops the AGENTS.md exclusion. That *un-retires* the path, which is the direction needing **no controlled migration**, so dev recovers on an ordinary deploy. Diff is exactly one line.
- **`bootstrapMaxChars` 32000 → 64000.** SOUL.md is 32,991 with the rules plus the new capability-denial guidance, so the cap had to move. It's our own number, not a platform limit — the schema types it as a plain integer with a documented default of 20000. At ~33k chars SOUL.md is ~8,250 tokens, **~4.6% of our 180k `contextTokens`**, sitting in the cached prefix where `cacheRead` is **10× cheaper** than `input`. Total usage 36,166 / 80,000.
- **`workspace_sync.py`'s comment** now records what was learned rather than the wrong conclusion.

## Kept from the split

Capability-denial rule, `psd-workflows` classified/certificated vocabulary, Slides + Sheets `--json-file` examples, the upstream retry, and the toolConfig canary. Only the file layout is reverted.

## Two rules now written into the code

1. **Verify a file is actually injected** (`openclaw config schema`) before building on it. A grep count proves the host mentions a name, not that it loads the file.
2. **Add a sync exclusion in the SAME release that first writes an image-owned file** — excluding a path already in S3 is a retired path and needs a migration; excluding one never pushed is free.

## Tests — mutation-checked

Moving the rules out of SOUL.md fails both: the rules body is measured inside SOUL.md, AGENTS.md is never treated as a bootstrap file, and the Dockerfile is asserted to append the body to SOUL.md.

## Verification

- agent-image Python suite as CI invokes it — **722 tests**
- Full unit suite; bootstrap-budget and config-consistency gates
- E2E green via the pre-push gate

## Deploy

**Ordinary deploy — no cutover.** No change to `storage-broker.ts` or migration 171, and the `workspace_sync.py` edit is comment-only. Removing an exclusion is the safe direction.

### DIFF
--- a/infra/agent-image/Dockerfile
+++ b/infra/agent-image/Dockerfile
@@ -294,31 +294,22 @@ COPY ${OPENCLAW_CONFIG} /home/node/.openclaw/openclaw.json
 # visible to the LLM. Keep roadmap notes here in the Dockerfile instead.
 # TODO(phase-2): Add Google Calendar and Gmail read-only tool access to the agent.
 #
-# Tier-1 enforcement: the rules in psd-rules SKILL.md are NOT auto-injected by
-# OpenClaw — the model only sees a Tier-2 catalog stub for any skill it has not
-# explicitly loaded. That meant the "non-negotiable" rules were never actually
-# present in the system prompt and the agent ignored them (observed 2026-05-03).
-# At build time we strip psd-rules' YAML frontmatter and write the body into a
-# bootstrap file so the rules ARE in the system prompt every turn. The canonical
-# edit point stays `skills/psd-rules/SKILL.md`.
-#
-# The rules go in AGENTS.md, NOT appended to SOUL.md (changed 2026-08-06).
-# OpenClaw budgets bootstrap files PER FILE (bootstrapMaxChars) as well as in
-# total (bootstrapTotalMaxChars), and it loads AGENTS.md, TOOLS.md, HEARTBEAT.md
-# and BOOTSTRAP.md alongside SOUL.md whenever they exist. Concatenating the
-# ~22k-char rules body onto SOUL.md spent one file's entire 32k budget while the
-# combined total sat at 35k of 80k — so we were out of room in one file while
-# 45k went unused, and every rule addition had to be paid for by deleting
-# another. Splitting also matches the documented roles: AGENTS.md is operational
-# rules and routing, SOUL.md is identity and voice.
+# Tier-1 enforcement: SOUL.md alone is auto-loaded, but the rules in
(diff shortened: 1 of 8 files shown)


krishagel:
@codex review

Head is `74d86b7ea7`. Your P1 is fixed.

You were right and it was worse than a stale comment: the Dockerfile revert put the `psd-rules` body back into SOUL.md, but SOUL.md's own Tier-0 line still said the rules were written to `AGENTS.md` and instructed the agent to "read `AGENTS.md`, not the end of this file." The rules sit directly below that sentence, so it would have sent the agent away from them to a file this host never injects — reproducing the exact symptom this PR exists to fix. SOUL.md now has zero AGENTS.md references.

Verified after the fix: SOUL.md 32,920 chars (under the raised 64,000 cap, total 36,095/80,000), bootstrap-budget and config-consistency gates pass, E2E green via the pre-push gate.

PR #1620, opened 2026-08-07
fix(agent-image): stop charging the prompt budget for files this host never injects

krishagel:
Codex P2 on #1619 — I merged that PR before reading this finding. It's a real consequence, not cosmetic.

## The problem

`BOOTSTRAP_FILES` drives **both** the build gate and the **runtime** check, and it listed `AGENTS.md`, `TOOLS.md` and `BOOTSTRAP.md` on the strength of upstream documentation. This host injects none of them — `openclaw config schema` enumerates the optional bootstrap files as SOUL/USER/IDENTITY/HEARTBEAT, and the injected-file arrays in `/app/dist` are `["SOUL.md"]` and `["IDENTITY.md","USER.md","SOUL.md"]`.

That became live as of #1619: the 2026-08-07 image pushed `AGENTS.md` into the S3 workspace of **every user who ran it**, and un-excluding the path means those copies get pulled back down. The runtime check would then find ~22k of dead weight on disk, charge it against the 80k total, and emit `BootTruncationWarn` + `BOOT_TRUNCATION` for a file with **no effect on the prompt at all**.

**Why my test missed it:** it only passed because the temporary directory happened to contain no `AGENTS.md`. It exercised the build path and never the runtime one — which is precisely where the stale file shows up.

## The fix

`BOOTSTRAP_FILES` is now the verified injected set: SOUL, IDENTITY, USER, MEMORY, HEARTBEAT.

Two tests, **mutation-checked** (re-adding `AGENTS.md` fails three):
- a 22,000-char stale `AGENTS.md` on disk is ignored by `runtime_bootstrap_sizes` and contributes nothing to the total
- the file set matches the host's injected list exactly, so nothing can be re-added on documentation alone

## Root cause, third time today

Trusting docs over `openclaw config schema`. Same error as `bootstrapPromptTruncationWarning` (host rejected the key) and the AGENTS.md split itself. The schema ships in the image and answers in seconds — the second test now encodes that so it can't drift back.

## Verification

- agent-image budget/config/wrapper/workspace/candidate suites — **236 tests**
- bootstrap-budget gate passing at **36,095 / 80,000**
- E2E green via the pre-push gate

Ordinary deploy — no cutover (no `storage-broker.ts`, no `workspace_sync.py`, no migration 171).

### DIFF
--- a/infra/agent-image/check_bootstrap_budget.py
+++ b/infra/agent-image/check_bootstrap_budget.py
@@ -39,18 +39,30 @@ import sys
 from typing import Dict, List, Optional, Tuple
 
 # The bootstrap files OpenClaw auto-loads when present (Dockerfile lines ~326-354
-# document SOUL/IDENTITY/USER/MEMORY as seeded, and AGENTS/TOOLS/HEARTBEAT/
-# BOOTSTRAP as also-auto-loaded when present). SOUL.md is the only one the image
-# builds by concatenation; the rest are copied/seeded as-is.
+# The files THIS host actually injects into the system prompt, verified against
+# `openclaw config schema` rather than documentation: skipOptionalBootstrapFiles
+# enumerates SOUL/USER/IDENTITY/HEARTBEAT, and the injected-file arrays in
+# /app/dist are ["SOUL.md"] and ["IDENTITY.md","USER.md","SOUL.md"]. SOUL.md is
+# the only one the image builds by concatenation; the rest are seeded as-is.
+#
+# AGENTS.md is deliberately ABSENT (2026-08-07). This host uses it for
+# post-compaction re-injection of named sections
+# (DEFAULT_POST_COMPACTION_SECTIONS = ["Session Startup", "Red Lines"]), not as
+# a per-turn bootstrap file. Counting it here charged the prompt budget for
+# something never injected — and because a stale AGENTS.md now exists in the S3
+# workspace of every user who ran the 2026-08-07 image, the runtime check would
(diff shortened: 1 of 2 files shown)


krishagel:
Closing — the premise is disproven, and @codex was right.

I ran the runtime experiment I should have run before making the claim: booted the image with distinct sentinels in `SOUL.md` and `AGENTS.md`, pointed a stub provider at a local capture server, and read the assembled request.

**Both sentinels appear in the `role=system` message**, with AGENTS.md injected under its own header:

```
## /home/node/.openclaw/AGENTS.md
# Rules

ZZAGENTSSENTINELZZ
```

So AGENTS.md **is** a main-session bootstrap file in the host we pin, and this PR would have undercounted the prompt and hidden real truncation — exactly the failure mode you described.

Your correction was precise on both points and I discounted both: `skipOptionalBootstrapFiles` enumerates only the *optional* files rather than the complete injected set, and the arrays I quoted from `/app/dist` are reduced-context allowlists, not the main-session loader. I also mis-cited `register-*.js` — that hit is the policy plugin's doctor check (`unmigratedToolsFinding`), not prompt assembly, so it neither supported nor refuted the point.

What this also means: the original diagnosis on #1616 was correct. `workspace_sync` overwriting AGENTS.md was the real bug and excluding it was the right fix — it failed only on ordering, because the first image pushed AGENTS.md to S3 before the exclusion existed, creating a retired path.

`dev` stays as merged in #1619: the rules are back in SOUL.md, which is also injected and demonstrably works. Leaving the split alone for now.

